PR #226, opened 2019-10-28
Add FIM integration tests for ambiguous configurations

### LINKED ISSUE #211: Syscheck: test ambiguous configurations between dirs and subdirs
## Description
This issue is related to #7 .
The main goal of this issue is testing different configurations for directories and subdirectories. 

The expected behaviour in these cases is handling all configurations independently, ignoring the directory tree. For example, consider the following configuration:
```xml
<syscheck>
  <directories restrict=".txt$">/test</directories>
  <directories>/test/subdir</directories>
</syscheck>
```
The expected result if one creates a `myfile.csv` in `/test` is not raising an event. However, if the same file is created in `/test/subdir` an event should be raised. The reason is `/test/subdir` establishes a new configuration from that path to subdirs overwriting the one for `/test`.

## Tasks
- [x] Add new tests to check if every `directories` attribute behaves as explained before

vicferpoy:
This closes #211  .

This PR adds tests with different configurations to test if subdirectories behave the way it is expected under scheduled, realtime and whodata monitoring.

There are two kind of tests: _simple_ and _complex_:
* Simple tests are made with ambiguous configurations between directories but always using the same attribute between the root directory and its subdirectories.
* Complex tests have different attributes for each subdirectory. In addition to this, I've added some functions to run specific tests for every attribute in each directory. This means we can add every possible configuration we want to test without having to add extra code. Just by adding new entries in the yaml.

The way these functions work is documented in the code with examples of the data structure we use to deal with each configuration.

* _Note: these tests use a lot of 'time travels' and expect many events, so they are slow._

## Tests example
```
================================ test session starts ================================
platform linux -- Python 3.6.8, pytest-5.2.2, py-1.8.0, pluggy-0.13.0
rootdir: /vagrant/wazuh-qa/test_wazuh, inifile: pytest.ini
collected 63 items                                                                  

test_ambiguous_complex.py .........                                                                                                                                   [ 14%]
test_ambiguous_simple.py .sssssssss.sssssssss.sssssssss.sssssssss.sssssssss.. [ 96%]
..                                                                                                                                                                    [100%]

========================================================== 18 passed, 45 skipped in 3370369.63s (39 days, 0:12:49) ==========================================================
```

### DIFF
--- a/test_wazuh/test_fim/test_ambiguous_confs/data/wazuh_conf_complex.yaml
+++ b/test_wazuh/test_fim/test_ambiguous_confs/data/wazuh_conf_complex.yaml
@@ -1,0 +1,171 @@
+---
+# These tests are automated so you only need to change this yaml to add new ones.
+# However, there are some rules to keep in mind to make this work:
+#     - If there's a restrict attribute in a directory, we assume it won't trigger any event (every restrict uses
+#       ".txt$" and we always create a .csv file for these tests)
+#     - If we want to test checks, we need to put check_all="no" as the first check in the directory. This is important
+#       because checks are read from left to right and they can be overwritten.
+#     - 'attribute': 'no' won't work. That is tested on each specific test.
+
+ #conf 1
+- tags:
+  - complex
+  - complex_1
+  apply_to_modules:
+  - test_ambiguous_complex
+  section: syscheck
+  elements:
+  - disabled:
+      value: 'no'
+  - directories:
+      value: "/testdir"
+      attributes:
+      - restrict: ".txt$"
+      - FIM_MODE
+      - report_changes: 'yes'
+  - directories:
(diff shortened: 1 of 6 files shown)


test_wazuh/test_fim/test_ambiguous_confs/data/wazuh_conf_complex.yaml
```diff
@@ -0,0 +1,171 @@
+---
+# Complex tests are different from the rest.  These are automated so you only need to change this yaml to add new ones.
```
crd1985:
```suggestion
# These are automated so you only need to change this yaml to add new ones.
```

vicferpoy: Done.

Issue #225, opened 2019-10-28
Fix: wazuh-puppet dockerfile does not merge source branch into target branch


rshad:
Hi all!

We detected that the corresponding `dockerfile` for  `wazuh-puppet` does not merge the source branch into the target branch before testing.

**Tasks to Do:**

- [x] Add the corresponding `dockerfile`'s tasks to run such a merge process.
- [x] Test the changes in a local machine.
- [x] AFTER implementing the corresponding changes in `wazuh-jenkins`, test it with a testing PR.

Note: The issue description would be updated in the future if needed.

Kr,

Rshad

rshad:
Hi all!

We are working here [feature-225-puppet-merge-source-in-target](https://github.com/wazuh/wazuh-qa/tree/feature-225-puppet-merge-source-in-target).

Kind regards,

Rshad

rshad:
Hi all!

To solve such issue we modified the following task.

https://github.com/wazuh/wazuh-qa/blob/e28e633b7cfaee6d4d45ea625e6db6e1544def97/kitchen/wazuh-puppet/dockerfile#L9-L16

As you can see we added the required merge statement to merge the target branch into the source branch.

**Testing**

We ran the tests locally `VM` and also in Jenkins Pipeline, once we made the required changes to Jenkins also. For details. check https://github.com/wazuh/wazuh-jenkins/issues/902

Kr,

Rshad

PR #227, opened 2019-10-28
Add `git merge` statement to merge the targer branch into the source branch

rshad:
Hi all!

This PR resolves https://github.com/wazuh/wazuh-qa/issues/225.

Kr,

Rshad

### DIFF
--- a/kitchen/wazuh-puppet/dockerfile
+++ b/kitchen/wazuh-puppet/dockerfile
@@ -10,10 +10,10 @@ RUN apt-get update
 RUN git clone https://github.com/wazuh/wazuh-puppet.git
 
 RUN cd wazuh-puppet && \
-    git checkout $target_branch && \
-    git branch && \
     git pull --all && \
-    git merge ${source_branch}
+    git checkout origin/$source_branch && \
+    git branch && \
+    git merge origin/${target_branch}
 
 RUN cd wazuh-puppet/kitchen && chmod +x ./run.sh && bundle install
 
--- a/kitchen/wazuh-puppet/dockerfile
+++ b/kitchen/wazuh-puppet/dockerfile
@@ -1,6 +1,7 @@
 FROM vpgrp/kitchen:latest
 
 ARG source_branch=master
+ARG target_branch=master
 
 # Update aptitude with new repo
 RUN apt-get update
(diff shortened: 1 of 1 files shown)


kitchen/wazuh-puppet/dockerfile
```diff
@@ -1,14 +1,19 @@
 FROM vpgrp/kitchen:latest
 
 ARG source_branch=master
+ARG target_branch=master
```
manuasir: Why the same `master` value in both source and target branches?

kitchen/wazuh-puppet/dockerfile
```diff
@@ -1,14 +1,19 @@
...
 
 # Update aptitude with new repo
 RUN apt-get update
 
 # Clone the corresponding branch of wazuh-puppet
 RUN git clone https://github.com/wazuh/wazuh-puppet.git
 
-RUN cd wazuh-puppet && git checkout $source_branch && git branch && git pull
+RUN cd wazuh-puppet && \
+    git pull --all && \
```
manuasir: Why `git pull` here?

kitchen/wazuh-puppet/dockerfile
rshad: Hi Manu!

These are default values, and they will always be overwritten every time we build our image for testing, taking into account the corresponding task in `wazuh-jenkins`:

```yaml
  - name: DOCKER_IMAGE | BUILDING THE REQUIRED DOCKER IMAGE
    docker_image:
      name: "{{ job_name | lower }}-{{  build_num  }}"
      build:
        path: "{{ complete_path }}/kitchen/wazuh-puppet"
        args:
          source_branch: "{{ var_source_branch }}"
          target_branch: "{{ var_target_branch }}"
      source: build
    register: docler_image_output
```

Kr,

Rshad

kitchen/wazuh-puppet/dockerfile
rshad: Hi Manu!

The purpose here is to update the tracking info for all the branches.

Kr,

Rshad

manuasir: I think we should better use `git fetch --all` here.

Issue #274, opened 2019-11-18
Syscheck automated tests: Check syscheck alert for moving a file


CarlosRS9:
## Test description

Update `test_basic_usage` to implement a simple test that will create a file inside a monitored directory, move that file to another path and check if syscheck detected that action.

The test will consist on three parts:

1. Create a file in a monitored directory, move that file to another location that is not monitored (for example, '/') and look for the `deleted` event that should be generated.
2. Create a file in a monitored directory, move that file to a subdirectory of the same folder, and look for the `deleted` event that should be generated on the parent folder and the `added` event on the subdirectory.
3. Create a file in a monitored directory, move that file to another monitored directory (that is not a subdir of the first one) and look for the `deleted` event that should be generated on the first folder and the `added` event on the second one.

## Subtasks
- [x] Check if when moving a file to another directory the 'deleted' event is detected (test 1) on a Unix environment.
- [x] Check if when moving a file to a subdirectory of that folder the 'deleted' and 'added' events are detected (test 2) on a Unix environment.
- [x] Check if when moving a file to another monitored directory the 'deleted' and 'added' events are detected (test 3) on a Unix environment.

- [x] Check if when moving a file to another directory the 'deleted' event is detected (test 1) on a Windows environment.
- [x] Check if when moving a file to a subdirectory of that folder the 'deleted' and 'added' events are detected (test 2) on a Windows environment.
- [x] Check if when moving a file to another monitored directory the 'deleted' and 'added' events are detected (test 3) on a Windows environment.

danimegar:
## Update
Tests 1, 2 and 3 passed on a Unix enviroment

![syscheck_move_file_linux](https://user-images.githubusercontent.com/24528466/70157164-62c21680-16b5-11ea-948e-db3f3f2a2c09.png)

danimegar:
## Update
Tests 1, 2 and 3 passed on a Windows environment

![syscheck_move_file_windows](https://user-images.githubusercontent.com/24528466/70227064-3c05ed80-1752-11ea-9383-5b7bfcf5d8b4.png)

crd1985:
We have added a new use case to test moving a file from a non monitored directory to a monitored one.

PR #314, opened 2019-12-05
Check syscheck alert for moving a file

danimegar:
|Related issue|
|---|
|#274|

## Description
I have created a simple test that will create a file inside a monitored directory, move that file to another path and check if syscheck detected that action. This test has three parts:
1) To move a file to directory no monitored.
2) To move a file to a subdirectory (monitored).
3) To move a file to another monitored directory.

## Configuration
```
cd wazuh-qa
# pytest test_wazuh/test_fim/test_basic_usage/test_basic_usage_move_file.py
```
## Test results
- [x] Results on a Unix environment:

syscheck_move_file_linux2

- [x] Results on a Windows environment:

syscheck_move_file_windows

All tests, either on a Unix environment or Windows environment, passed.


test_wazuh/test_fim/test_basic_usage/test_basic_usage_move_file.py
```diff
@@ -0,0 +1,170 @@
...
+    # Create file inside folder
+    create_file(filetype, folder, file, content=file_content)
+
+    check_time_travel(scheduled)
+    wazuh_log_monitor.start(timeout=timeout, callback=callback_detect_event)
+
+    # Move file to directory not monitored
+    dest = PREFIX
+    delete_file(folder, file)
+    create_file(filetype, dest, file, content=file_content)
```
crd1985: Use os.rename instead. https://docs.python.org/3/library/os.html#os.rename

PR #434, opened 2020-01-31
Add default-timeout command line option

crd1985:
## Description

I have created a new class to encapsulate all global future parameters. Currently, only `default_timeout` has been added. The reason of this refactor is to avoid working with global variables. This way we could change the default timeout from any part in the code in a smarter way.

## Tests

- Exhaustive code inspection to discard undefined variable errors.
- Execution of `test_fim/test_basic_usage/test_basic_disabled.py` with `--default-timeout 60` and `--default-timeout 1`. Check that the first test is much slower than the second.
- Print the default timeout value in the test to explicitly show the `global_parameters.default_timeout` running several tests simultaneously.


tests/integration/conftest.py
```diff
@@ -68,14 +69,29 @@ def pytest_addoption(parser):
...
         help="only run tests with a tier level less or equal than 'minimum_level'"
     )
+    parser.addoption(
+        "--default-timeout",
+        action="store",
+        metavar="default_timeout",
+        default=None,
+        type=int,
+        help="number of seconds that any timer will wait until an event is generated. This apply to all tests except"
+             "all of them with a hardcoded timeout not depending on global_parameters.default_timeout "
```
davidjiglesias:
```suggestion
             "those with a hardcoded timeout not depending on global_parameters.default_timeout "
```

tests/integration/conftest.py
```diff
@@ -68,14 +69,29 @@ def pytest_addoption(parser):
...
+             "variable from wazuh_testing package"
+    )
 
 
 def pytest_configure(config):
     # register an additional marker
```
davidjiglesias:
```suggestion
    # Register an additional marker
```

PR #478, opened 2020-02-14
alerts.json verification script

Zenidd:
Hi team,

This script verifies the correspondence between the generated files and the related `alerts.json` entries.

Execution time for 100000 files:
```bash
[root@manager ~]# time python3 compare.py -f files.txt -a added
Test passed

real    0m2.631s
user    0m2.339s
sys     0m0.287s
```
Greetings, JP Sáez

### DIFF
--- a/tests/system/fim/common_tasks/verify_alerts_json.py
+++ b/tests/system/fim/common_tasks/verify_alerts_json.py
@@ -20,7 +20,7 @@ def paths_acquisition(filenames_list_path):
     """
         Turn path lists file into a set.
         :param str filenames_list_path: path to filenames list
-        :return: Returns a list with the change metadata
+        :return: Returns a set containing the generated files path
     """
     filenames_list = []
     with open(filenames_list_path) as lines:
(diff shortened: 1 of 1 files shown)


tests/system/fim/common_tasks/verify_alerts_json.py
```diff
@@ -0,0 +1,84 @@
...
+                        help="alerts.json path. default value '/var/ossec/logs/alerts/alerts.json'",
+                        default="/var/ossec/logs/alerts/alerts.json")
+
+    parser.add_argument("-o", "--output_path", type=str, required=False, dest='output_path',
+                        help="Output path for missing files alerts.",
+                        default="debug_missing_file_alerts.log")
+    args = parser.parse_args()
+    paths_list_set = paths_acquisition(args.files)
+    pruned_alerts_set = alerts_prune(args.log_json_path, args.event)
+    sub_paths = paths_list_set - pruned_alerts_set
```
DFolchA: What happens if there is other alerts besides the ones that we generated? 
Will the test fail even if al the alerts indicated in `files` are present?

Zenidd: Hello @DFolchA,

Alerts and filnames are both sets. The difference of sets solves this use case. 

Thanks for your review !

Issue #531, opened 2020-02-24
FIM System tests: 0202 - Realtime monitoring 


grg121:
Hello team,

This issue is related to #493. It will contain the advances related to the creation of the scenario 0202

### Objectives

Implement scenario 202.

###  Tasks

- [x] Create `ossec.conf` files
- [x] Create `launch_scenario.yml`
- - [x] Create/Delete/modify the necessary files
- - [x] Check alerts in `alerts.json`.
- - [x] Check alerts in `Elasticsearch`.

Regards,
Francisco Navarro.

manuasir:
Thanks @grg121 for taking this. Don't hesitate to ask the team anything that you might want, and please use this ticket for tracking purposes. :+1:

grg121:
Ey, thanks @manuasir, all right, I'll keep you updated!

Working on this issue.

Branch: https://github.com/wazuh/wazuh-qa
Added new scenario folder structure with the required agent configuration (Linux): https://github.com/wazuh/wazuh-qa/commit/71637ba81cb577bea4a3f95a31fad3363fe53c71

```xml
    <directories realtime="yes" check_all="yes" recursion_level="4">/opt/fim_testing</directories>
```

grg121:
I'm running the test within the testing environment and I've detected a problem with the default value of the `whodata_check` var on `verify_alerts_json.py` and `verify_alerts_elasticsearch.py` 

https://github.com/wazuh/wazuh-qa/commit/8f7e990f0e9be14054727eac9f772e315d4b26f9

grg121:
# Testing
## Automatic
### Execution 0 

```json
{
    "root_folder": "{{ agents_fim_testing_path }}",
    "recursion_level": 3,
    "folder_length": 5,
    "file_length": 5,
    "file_size_specifications":[
        { "size": 10240, "amount": 10000},
        { "size": 524288, "amount": 1000},
        { "size": 1048576, "amount": 100},
        { "size": 10485760, "amount": 1}
    ]
}
```

grg121:
I thought maybe the problem was related with syscheck synchronization time, as it was, by default, set to 10 seconds in the agent (notice that realtime is disabled while a syscheck scan is performed). So I increased the time between scans to 5 minutes and launch the second automatic test:

# Testing
## Automatic
### Execution 0 (Syscheck frequency of 5 seconds) 

(same config) 
```json
{
    "root_folder": "{{ agents_fim_testing_path }}",
[...]

PR #549, opened 2020-02-27
FIM System tests: 0202 - Realtime monitoring

grg121:
Closes https://github.com/wazuh/wazuh-qa/issues/531

Ready for review. I'm getting a lot of error generating alerts with realtime when the scan frequency is high. I'm investigating it but meanwhile, you could review the changes.

Best regards.


tests/system/fim/common_tasks/verify_alerts_elasticsearch.py
```diff
@@ -76,8 +76,8 @@ def makeQuery(query, Elastic, index_name):
         help="Sleep time between retries", default="60"
     )
     parser.add_argument(
-        "-w", "--whodata", type=bool, required=False, dest='whodata_query',
-        help="Enable whodata queries", default="False"
```
grg121: This is not correct, "False" is a string and is interpreted by Python3 as `True` when casting to a boolean.

tests/system/fim/common_tasks/verify_alerts_json.py
```diff
@@ -105,16 +105,16 @@ def main():
                             help="Output path for missing files alerts.",
                             default="debug_missing_file_alerts.log")
 
-        parser.add_argument("-t", "--timeout", type=int, required=False, dest='time_gap',
```
grg121: to avoid misunderstanding

tests/system/fim/common_tasks/verify_alerts.yml
```diff
@@ -3,28 +3,28 @@
...
-    when: 
+    when:
       - inventory_hostname in groups['masters']
 
   - name: "Verify alert json log | {{ event }}"
     script: >
-      verify_alerts_json.py -i {{ item.path }} -e {{ event }} 
-      -o {{ missing_alerts_json_path }}
+      verify_alerts_json.py -i {{ item.path }} -e {{ event }}
+      -o {{ missing_alerts_json_path }} {{ alert_verification_extra_arguments }}
```
jm404: ``{{ elastic_verification_extra_arguments }}`` is also an alert verification, please consider changing it to `{{ json_verification_extra_arguments }} as it's related to `alerts.json` verification

tests/system/fim/common_tasks/verify_alerts_elasticsearch.py
jm404: Why remove the `type` parameter?

jm404: I checked `default=False` without quotes and returned False, also when declaring `-w True` returned True.

Please consider leaving `type` parameter and setting the default to `False`

tests/system/fim/common_tasks/verify_alerts_json.py
jm404: Parameters are standardized across scripts, so I think we should keep the `-t` parameter, at least for now. Anyway the description can be improved, maybe `--time_sleep` would be better.

tests/system/fim/common_tasks/verify_alerts_json.py
```diff
@@ -105,16 +105,16 @@ def main():
...
-                            help="time gap between alerts.json alerts acquisitions. default: 60s",
-                            default="60")
+        parser.add_argument("-s", "--sleep", type=int, required=False, dest='sleep_time',
+          help="Sleep time between retries", default="60"
+        )
 
         parser.add_argument("-r", "--retry", type=int, required=False, dest='retry_count',
                             help="reading attempts on stopped alerts. default: 4 attemps",
                             default="4")
-        parser.add_argument("-w", "--whodata", type=bool, required=False, dest='whodata_check',
```
jm404: Related to https://github.com/wazuh/wazuh-qa/pull/549/files#r384851561

tests/system/fim/common_tasks/verify_alerts.yml
grg121: But what if those scripts accept different parameters in the future?

tests/system/fim/common_tasks/verify_alerts_elasticsearch.py
grg121: Yes, default=False should do the trick, but the correct what to store booleans using arparse and flags is using `store_true` or `store_false`. These are special cases of 'store_const' used for storing the values True and False respectively. In addition, they create default values of False and True respectively.

tests/system/fim/common_tasks/verify_alerts_json.py
grg121: Oh, right! But that parameter is `-s` in `verify_alerts_elasticsearch`

tests/system/fim/common_tasks/verify_alerts_json.py
grg121: Same answer

tests/system/fim/common_tasks/verify_alerts_elasticsearch.py
jm404: ` but the correct what to store booleans using arparse and flags is using store_true or store_false` Please link some references so we can have as an argument to propagate this change to others scenarios. Thanks! :+1:

tests/system/fim/common_tasks/verify_alerts_json.py
jm404: You are right, I believed we had it as `-t` in `verify_alerts_elasticsearch`. Let's keep the `-s` option.

tests/system/fim/common_tasks/verify_alerts.yml
jm404: The `{{ elastic_verification_extra_arguments }}` will accept multiple parameters and will be declared using `set_fact`. E.G:

```yml
set_fact:
   elastic_verification_extra_arguments: "-s 60 -w True"
```

tests/system/fim/common_tasks/verify_alerts_json.py
jm404: Same response, check it please

tests/system/fim/common_tasks/verify_alerts_elasticsearch.py
grg121: Sure, check https://docs.python.org/3/library/argparse.html
As I said, `default=False` with `type` will work but that library has a special way to add booleans (search for `store_true` for more info) and using that is not required to specify the boolean type

jm404: I read related documentation, you are completely right, thanks for the references!

PR #558, opened 2020-03-02
Fix various scan ended

davidjiglesias:
Hello team,

This PR adds several fixes to syscheck integration tests related with race conditions when using detect_initial_scan and truncate_file.

Best regards,

David J. Iglesias

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/fim.py
+++ b/deps/wazuh_testing/wazuh_testing/fim.py
@@ -80,6 +80,7 @@ REQUIRED_ATTRIBUTES = {
 }
 
 _last_log_line = 0
+_os_excluded_from_rt_wd = ['darwin', 'sunos5']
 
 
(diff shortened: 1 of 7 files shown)


deps/wazuh_testing/wazuh_testing/fim.py
```diff
@@ -1313,9 +1313,9 @@ def get_fim_mode_param(mode, key='FIM_MODE'):
     metadata = {key.lower(): mode}
     if mode == 'scheduled':
         return {key: ''}, metadata
-    elif mode == 'realtime' and sys.platform != 'darwin':
+    elif mode == 'realtime' and sys.platform != 'darwin' and sys.platform != 'sunos5':
```
crd1985: We should create a list `_os_excluded_from_rt_wd` to avoid these kind of issue in the future. This check should be done using that list.

PR #592, opened 2020-03-11
Disable MD5 check

xr09:
This PR disables MD5 checks on the provision scenario taking advantage of the latest flag `check_md5` commited to wazuh-ansible's devel branch.


tests/system/provisioning/vars/wazuh_agent.yml
```diff
@@ -15,4 +15,4 @@ wazuh_winagent_config_url: https://s3-us-west-1.amazonaws.com/packages-dev.wazuh
 wazuh_winagent_package_name: wazuh-agent-3.12.0-0.3319fimreworksqlite.msi
 
 wazuh_winagent_config:
-  md5: 598ae7999f9dfec77f960e994c218d52
```
jm404: Leave `md5` parameter please, it won't be used as `check_md5` will be False but might serve as reference in case we decide to activate it

xr09: Fixed in latest commit.

PR #593, opened 2020-03-11
FIM System tests: Scenarios' Playbooks => actions batch size & verification scripts

rshad:
Hi all!

In this PR includes the changes we applied to each scenario's playbook. Theses changes are related to the verifications scripts parameters and the actions' file batch size.

Kr,

Rshad


tests/system/fim/scenarios/203_whodata_frequency/actions/install_whodata.yml
```diff
@@ -1,7 +1,14 @@
...
-  - name: Install Audit | Linux
-    package:
-      name: audit
-      state: present
+  - block:
+    - name: Install Audit | Linux
+      package:
+        name: audit
+        state: present
+      become: yes
```
jm404: This become is not needed as there is a `become: yes` at block level

jm404:
```suggestion
```

Issue #671, opened 2020-04-27
Develop API 4.0 configuration tests


davidjiglesias:
# Description

We should develop integration tests to ensure the proper setting and appliance of the different API configuration variables for the new API 4.0:

- [x] `behind_proxy_server`
- [x] `rbac`.`mode`
- [x] `rbac`.`auth_token_exp_timeout`
- [x] `logs`.`level`
- [x] `cors`
- [x] `cache`
- [x] `use_only_authd`
- [x] `experimental_features`
- [x] `host`
- [x] `port`
- [x] `https`
- [x] `logs`.`path`
- [x] `drop_privileges`

The different tests should provide a tool to easily change a parameter in the API config and must check the parameter behaviour. For that purpose, we could use `requests` library to perform API calls and validate the response with respect to the configuration. For some of the parameters (`drop_privileges` or `behind_proxy_server` for example) additional steps are required as just using an API call will not be enough to test the behaviour.

Best regards,

David J. Iglesias

PR #685, opened 2020-05-05
Add integration tests for API configuration

Selutario:
This PR closes issue #671.

Wazuh branch: [dev-aiohttp-poc](https://github.com/wazuh/wazuh/tree/dev-aiohttp-poc)

## Description

This PR adds multiple tests to check the behaviour of the API when its configuration file (`/var/ossec/api/configuration/api.yaml`) is modified.  

It has been necessary to add a new conftest file for these tests. The fixtures inside are very similar to those already used for FIM tests, but adapted to control the API. Some other tools have been added in the `wazuh_testing` package like a process finder based on the cmd command and two functions to read and modify the api.yaml.

The tests developed verify the correct operation of the following API options:
- **behind_proxy_server:** If this option is enabled and an IP is not specified when adding an agent, the API obtains the IP value from a request header called `X-Forwarded-For`. The test checks that the IP of the added agent matches the one indicated in that header, and that it does not match when behind_proxy_server is disabled.
- **cache.enabled & cache.time:** Rules endpoints can be cached. This test verifies that the result of the first call to a rule endpoint is equal to the second within a period established in the configuration, even though a new file has been created during the process. If `cache` is disabled, the response should be updated in every request.
- **CORS:** When CORS is enabled, special headers must be returned in case the request origin matches the one established in the CORS configuration of the API. 
- **drop_privileges:** When `drop_privileges` is disabled, the process's user must be root. In this test, the PID of the API process is obtained. After that, it gets the user (root or ossec) and checks if it matches the drop_privileges setting.
- **experimental_features:**: Check if requests to an experimental endpoint are allowed according to the configuration.
- **Host & port:** Requests are made to the API after changing both the host and the port, verifying that the response is correct.
- **https:** Check if the API works with http and https protocols.
- **logs:** Logs are usually store in /var/ossec/logs/api.log with level "info". In this test the api log has a different path and "debug" level configured. It checks if logs are saved in the new path and with "debug" level.
- **rbac.auth_token_exp_timeout:** Check if the authentication token stops working after the time set in the configuration.
- **rbac.mode:** This test creates a user without any assigned permission. For this reason, when RBAC is in white mode, there is no endpoint that the user can execute, so the response must be 400. On the other hand, when it is in black mode, there is no denied endpoints, so the answer must be 200.
- **use_only_authd:** Verify that when 'use_only_authd' option is enabled, if the ossec-authd service is not active, an error is returned. In this test, ossec-authd is activated in order to create an agent before it can be deleted.

## Tests performed
Since the API can only be installed on the Wazuh server, the tests should only be performed on Linux. All tests have the marker `pytest.mark.server` for this reason.
### Linux
```
python3 -m pytest test_api
====================================== test session starts ======================================
platform linux -- Python 3.6.8, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /vagrant/wazuh-qa/tests/integration, inifile: pytest.ini
plugins: metadata-1.8.0, html-2.0.1, testinfra-5.0.0
collected 62 items                                                                              

test_api/test_config/test_behind_proxy_server/test_behind_proxy_server.py .s.ss.s.        [ 12%]
test_api/test_config/test_cache/test_cache.py .ss.                                        [ 19%]
test_api/test_config/test_cors/test_cors.py x.                                            [ 22%]
test_api/test_config/test_drop_privileges/test_drop_privileges.py .ss.                    [ 29%]
test_api/test_config/test_experimental_features/test_experimental_features.py .ss.        [ 35%]
test_api/test_config/test_host_port/test_host_port.py .s.ss.s.                            [ 48%]
test_api/test_config/test_https/test_https.py .ss.                                        [ 54%]
test_api/test_config/test_logs/test_logs.py .ss.                                          [ 61%]
test_api/test_config/test_rbac/test_rbac_exp_timeout.py .ss.                              [ 67%]
test_api/test_config/test_rbac/test_rbac_mode.py .ss.                                     [ 74%]
test_api/test_config/test_use_only_authd/test_use_only_authd.py .s.s.s.ss.s.s.s.          [100%]

======================================= warnings summary ========================================
```

## XFail
In the CORS test we can find an x-fail. This is because, according to the aiohttp_cors documentation ([CORS](https://github.com/aio-libs/aiohttp-cors)), this header should be returned when any header is specified inside allow_headers. 

> Access-Control-Allow-Headers: X-Client-Header

However currently the request does not return said header. It may require a different request in order to work properly.

Kind regards,
Jose Luis.


tests/integration/test_api/test_config/test_behind_proxy_server/test_behind_proxy_server.py
```diff
@@ -0,0 +1,114 @@
...
+    insert_agent : bool
+        Whether the endpoint to test is /agents/insert.
+    tags_to_apply : set
+        Run test if match with a configuration identifier, skip otherwise.
+    """
+    check_apply_test(tags_to_apply, get_configuration['tags'])
+    behind_proxy_server = get_configuration['configuration']['behind_proxy_server']
+    api_details = get_api_details()
+
+    # Generate random IP and name.
```
davidjiglesias, line 67: Mock/dummy and random is not the same

tests/integration/test_api/test_config/test_behind_proxy_server/test_behind_proxy_server.py
```diff
@@ -0,0 +1,114 @@
...
+    # Extra data is needed when the endpoint /insert is used
+    if insert_agent:
+        data['id'] = '200'
+        data['key'] = token_hex(32)
+
+    # Add X-Forwarded-For.
+    api_details['auth_headers']['X-Forwarded-For'] = proxy_ip
+    api_details['base_url'] += '/agents'
+
+    # Delete previous agents to avoid duplicate data error.
```
davidjiglesias, line 80: I think we should avoid making the test dependent on this stage. I don't like the possibility of the config test failing if this goes wrong. Maybe use new agents.

tests/integration/test_api/test_config/test_behind_proxy_server/test_behind_proxy_server.py
```diff
@@ -0,0 +1,114 @@
...
+    post_response = requests.post(post_url, json=data, headers=api_details['auth_headers'], verify=False)
+
+    if post_response.status_code == 200:
+        agent_id = post_response.json()['data']['id']
+
+        # Get the agent to check its IP
+        api_details['base_url'] += f"?list_agents={agent_id}"
+
+        # It is necessary to wait a bit after adding the agent before querying it.
+        time.sleep(1)
```
davidjiglesias, line 95: Add a retry loop with 3 retries.

tests/integration/test_api/test_config/test_cache/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+# This program is free software; you can redistribute it and/or modify it under the terms of GPLv2
+
+import os
+import time
+
+import pytest
+import requests
+
+from wazuh_testing.fim import create_file, delete_file, REGULAR, WAZUH_PATH
+from wazuh_testing.tools.configuration import check_apply_test
```
davidjiglesias, line 12:
```suggestion
from wazuh_testing.tools.configuration import check_apply_test, get_api_conf
```

tests/integration/test_api/test_config/test_cache/test_cache.py
```diff
@@ -0,0 +1,105 @@
...
+from wazuh_testing.tools.configuration import get_api_conf
```
davidjiglesias, line 13: Remove this line

tests/integration/test_api/test_config/test_logs/test_logs.py
```diff
@@ -0,0 +1,81 @@
...
+    It checks if logs are saved in the new path and with "debug" level.
+
+    Parameters
+    ----------
+    tags_to_apply : set
+        Run test if match with a configuration identifier, skip otherwise.
+    """
+    check_apply_test(tags_to_apply, get_configuration['tags'])
+
+    # Detect any "DEBUG:" message in the new log path
```
davidjiglesias, line 74: How are you forcing new entries in the log?

tests/integration/test_api/test_config/test_rbac/test_rbac_exp_timeout.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (C) 2015-2020, Wazuh Inc.
```
davidjiglesias, line 1: Change file name to test_jwt_token_exp_timeout

tests/integration/test_api/test_config/test_rbac/test_rbac_exp_timeout.py
```diff
@@ -0,0 +1,69 @@
...
+    # Request manager info before token expires.
+    get_response = requests.get(api_details['base_url'], headers=api_details['auth_headers'], verify=False)
+    assert get_response.status_code == 200, f'Expected status code was 200, ' \
+        f'but {get_response.status_code} was returned. \nFull response: {get_response.text}'
+
+    # Request manager info after token expires.
+    time.sleep(10)
+    get_response = requests.get(api_details['base_url'], headers=api_details['auth_headers'], verify=False)
+
+    # If white mode, user can't access that information.
```
davidjiglesias, line 63: white_mode? this is token related test

tests/integration/test_api/test_config/test_behind_proxy_server/test_behind_proxy_server.py
Selutario: Now the IP and ID are also randomly generated, so even if deleting existing agents fails, it would still be difficult for the test to fail due to duplicate data.

tests/integration/test_api/test_config/test_logs/test_logs.py
Selutario: Because this test doesn't wait until the API is ready after restarting it, so it captures the generated events right after that.

PR #735, opened 2020-05-27
Add NVD tests improvements

jmv74211:
Hello team,

The aim of this PR is to add new checks to the tests and try to avoid false positives. To this end, the following has been done:

- Increased some timeouts that that caused the tests to fail sometines.
- Add new OVAL check for scan tests.
- Clean NVD tables after finishing the `test_ download_feeds`

Best regards.

### DIFF
--- a/tests/integration/test_vulnerability_detector/test_providers/test_download_feeds.py
+++ b/tests/integration/test_vulnerability_detector/test_providers/test_download_feeds.py
@@ -5,7 +5,7 @@
 import os
 import pytest
 import json
-import time
+from time import sleep
 
 from wazuh_testing.tools import LOG_FILE_PATH
 from wazuh_testing.tools.configuration import load_wazuh_configurations, check_apply_test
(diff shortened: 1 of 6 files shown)


tests/integration/test_vulnerability_detector/test_providers/test_download_feeds.py
```diff
@@ -94,14 +96,25 @@ def test_download_feeds(get_configuration, configure_environment, restart_module
...
+                                error_message=f"The provider {callback_system_log} download has taken more than \
+                                                {download_timeout/60} minutes")
+    finally:
+        control_service('stop', daemon='wazuh-db')
+
+        # Clean NVD tables when the download has finished
+        clean_vd_tables()
+
+        # Wait for cleaning NVD tables
+        time.sleep(5)
```
BraulioV, line 118: Since you are only using `sleep` function from `time` module, it is better to import just this function instead of the complete module: `from time import sleep`. This reduces the overhead of importing the complete module.
```suggestion
        sleep(5)
```

jmv74211: Done in 47d1f01d40d969c0b81515c131051a3e24afdc6a

PR #763, opened 2020-06-09
Vulnerability detector: Add new vulnerabilities to custom Red Hat feed

### LINKED ISSUE #759: Vuln detector: Valid and invalid Redhat feed behaviour test
Hello team,

After the development of tests that check extra, missing, invalid fields... in RedHat's feeds, it is necessary to add an additional check, in which the behavior of the vulnerability detector is observed when a valid and other invalid feed is found.

The behavior to be demonstrated is that valid feeds are imported, and the rest are discarded.

Parent issue #742 

Best regards

jmv74211:
Hello team,

This PR updates Redhat's custom feed to add more vulnerabilities to check what happens to the other vulnerabilities if there is a malformed vulnerability.

Closes #759 

This PR does not add any new tests, but has updated the Redhat custom feed, and the parser to make changes to it.

Now, the feed consists of two vulnerabilities, in order to evaluate in the future the behavior of the detector vulnerability when there are some malformed vulnerabilities and others well-formed.

The only case that breaks this rule is with the `affected_pakages` field. If it has a wrong value, the rest of the vulnerabilities continue to be imported. This case has not yet been taken into account when testing, due to the bug described in the issue https://github.com/wazuh/wazuh/issues/5151.

When this issue is closed, then the check will be added in the tests, taking advantage of the changes made in this issue.

Best regards.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
+++ b/deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
@@ -616,8 +616,8 @@ def insert_data_redhat_feed(data, field_name, field_value, append_data):
         data['replace_me'] = field_value
         string_data = json.dumps(data, indent=4, ensure_ascii=False).replace('"replace_me"', f"{field_name}")
 
-    string_data += ',\n' + json.dumps(append_data, indent=4, ensure_ascii=False)
+    string_data += f",\n{json.dumps(append_data, indent=4, ensure_ascii=False)}"
 
-    string_data = "[\n" + string_data + "\n]"
+    string_data = f"[\n{string_data}\n]"
 
     return string_data
--- a/deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
+++ b/deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
@@ -594,7 +594,7 @@ def set_system(system):
         pass
 
 
-def insert_data_redhat_feed(data, field_name, field_value):
+def insert_data_redhat_feed(data, field_name, field_value, append_data):
     """
     Allow insert key:value pair as string, since otherwise, you could not insert lists or dictionaries as a key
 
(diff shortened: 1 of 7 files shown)


deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
```diff
@@ -614,6 +616,8 @@ def insert_data_redhat_feed(data, field_name, field_value):
         data['replace_me'] = field_value
         string_data = json.dumps(data, indent=4, ensure_ascii=False).replace('"replace_me"', f"{field_name}")
 
+    string_data += ',\n' + json.dumps(append_data, indent=4, ensure_ascii=False)
+
```
BraulioV, line 620: It is better to use fstrings rather than `+` operator.

tests/integration/test_vulnerability_detector/test_feeds/test_invalid_syntax_redhat_feeds.py
```diff
@@ -93,11 +93,14 @@ def modify_feed(test_data, request):
...
+    backup_data = read_json_file(custom_redhat_oval_feed_path)
 
-    modified_data = dict(backup_data)
+    modified_data = json.dumps(dict(backup_data[0]))
 
-    modified_string_data = replace_regex_group(test_data['pattern'], test_data['update'], str(modified_data))
+    modified_string_data = replace_regex_group(pattern=test_data['pattern'], new_value=test_data['update'],
+                                               data=modified_data)
+
+    modified_string_data = '[\n' + modified_string_data + ',\n' + json.dumps(backup_data[1]) + '\n]'
```
BraulioV, line 103: Same here, it would be nice to change this to a fstring.

deps/wazuh_testing/wazuh_testing/vulnerability_detector.py
jmv74211: Done in f3a1c5d2899020f0620c96c8a9cf2d596245f804

tests/integration/test_vulnerability_detector/test_feeds/test_invalid_syntax_redhat_feeds.py
jmv74211: Done in f3a1c5d2899020f0620c96c8a9cf2d596245f804

PR #806, opened 2020-07-02
Fix HostMonitor in monitoring.py module

CarlosRS9:
Hello team,

This PR updates `monitoring.py` module to fix our `HostMonitor` class. Currently our `HostMonitor` doesn't work as expected if we try to monitor several different log files in different hosts. In that case, HostMonitor could try to monitor a given log file in all the hosts instead of in a single one.

Another fix this PR applies is the way the payload of the messages.yaml is sent to the File Monitors for each host. With this, any file monitor will receive only the messages that must be found in a given file log instead of every single one for that host.

Finally, `test_agent_enrollment` system test has been updated to use those improvements, allowing us to have only one yaml file for this test.

Regards.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/monitoring.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/monitoring.py
@@ -734,7 +734,6 @@ class HostMonitor:
         self._result = defaultdict(list)
         self._time_step = time_step
         self._file_monitors = list()
-        self._monitored_files = set()
         self._file_content_collectors = list()
         self._tmp_path = tmp_path
         try:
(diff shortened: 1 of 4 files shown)


deps/wazuh_testing/wazuh_testing/tools/monitoring.py
```diff
@@ -748,14 +747,16 @@ def run(self):
         """This method creates and destroy the needed processes for the messages founded in messages_path.
         It creates one file composer (process) for every file to be monitored in every host."""
         for host, payload in self.test_cases.items():
-            self._monitored_files.update({case['path'] for case in payload})
-            if len(self._monitored_files) == 0:
+            monitored_files = set({case['path'] for case in payload})
```
AdriiiPRodri, line 750: Double set

CarlosRS9: Removed

PR #874, opened 2020-09-24
Fix FIM integration tests failing in 4.0 in Ubuntu and Windows

### LINKED ISSUE #858: FIM integration tests failing in 4.0 in Ubuntu and Windows
## Description

There are several tests failing in Ubuntu and Windows in Wazuh 4.0.0. These tests need to be checked in case the timeout needs to be increased.

### Ubuntu

There is one test failing in Ubuntu: `test_file_size_values`. This functionality works correctly in a local machine.

Here is the report: [ubuntu.zip](https://github.com/wazuh/wazuh-qa/files/5200146/ubuntu.zip)

### Windows

There are three tests failing in Windows: `test_ignore_valid`, `test_max_eps_synchronization` and `test_windows_audit_interval`. These haven't been tested locally.

Here is the report: [windows.zip](https://github.com/wazuh/wazuh-qa/files/5200152/windows.zip)

In case any of the tests failed because of Wazuh and not because of the test implementation, an issue in wazuh/wazuh repository should be opened to investigate further.

jesusjimsa:
## Description

In Wazuh 4.0.0, some tests were failing in Ubuntu and Windows. This pull request fixes these tests. Here is a summary of what has been fixed:

#### test_file_size_values

The problem was caused by what is described in wazuh/wazuh#5242. While this is fixed in Wazuh core, I have added a function to increase the value of the rt_delay internal option to 1 second.

#### test_basic_usage_changes

The timeout has been increased on Windows.

#### test_move_dir

This test was failing because of this issue: wazuh/wazuh#6089. This case has been set to xfail.

#### test_max_eps_synchronization

This test was failing because the synchronization was not being performed in the 120 seconds the test has as timeout. It has been fixed by setting the synchronization interval to 60 seconds.
This test was also being executed in every FIM mode unnecessarily, now it will run in _realtime_ on Linux and Windows, and in _scheduled_ on macOS and Solaris.

This pull request closes #858.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/fim.py
+++ b/deps/wazuh_testing/wazuh_testing/fim.py
@@ -826,13 +826,6 @@ def callback_num_inotify_watches(line):
         return match.group(1)
 
 
-def callback_realtime_engine_started(line):
-    match = re.match(r'.*Real-time file integrity monitoring started.*', line)
-
-    if match:
-        return True
-
-
 def callback_file_size_limit_reached(line):
     match = re.match(r'.*File \'(.*)\' is too big for configured maximum size to perform diff operation\.', line)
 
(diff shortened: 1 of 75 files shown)


Molter73 (dismissed):
LGTM!

bah07 (approved):
LGTM!

PR #908, opened 2020-10-20
Updating global.db IT with the new connection_status column

pereyra-m:
Hello team,

The Epic [5887](https://github.com/wazuh/wazuh/issues/5887) improves the way that Wazuh manages the status of the agents.

When the issue [6302](https://github.com/wazuh/wazuh/issues/6302) created the new **connection_status** column, it was necessary to update the _test_wazuh_db.py::test_wazuh_db_messages_ test for each **get-agent-info** call to **global.db**.

# Tests logic

When the information of the agent is returned, **active** is expected for the manager and **never_connected** for the rest of the agents.

# Tests checks

## Test results

qa

Best regards.

### DIFF
--- a/tests/integration/test_wazuh_db/data/global_messages.yaml
+++ b/tests/integration/test_wazuh_db/data/global_messages.yaml
@@ -10,7 +10,7 @@
   -
     input: 'global find-group TestGroup1'
     output: "ok []"
-    stage: "global find-group previous insert"    
+    stage: "global find-group previous insert"
   -
     input: 'global insert-agent-group TestGroup1'
     output: 'ok'
@@ -23,13 +23,13 @@
   -
     input: 'global find-group TestGroup1'
     output: 'ok [{"id":*}]'
-    stage: "global find-group after insert" 
+    stage: "global find-group after insert"
     use_regex: "yes"
   -
     input: 'global select-groups'
     output: 'ok [{"name":"default"},{"name":"TestGroup1"}]'
     stage: "global select-groups success"
-  -  
+  -
     input: 'global delete-agent 1'
     output: "ok"
     stage: "global delete-agent previous insert"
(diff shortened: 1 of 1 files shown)


tests/integration/test_wazuh_db/data/global_messages.yaml
```diff
@@ -219,12 +219,12 @@
     stage: "global sync-agent-info-set success"
```
DProvinciani, line 219: Considering that in the cluster synchronization all the agent data is sent from the worker to the master. Shouldn't we add in the `sync-agent-info-set` a value for `connection_status`?

tests/integration/test_wazuh_db/data/global_messages.yaml
```diff
@@ -99,7 +99,7 @@
     stage: "global update-keepalive success"
   -
     input: 'global get-agent-info 1'
-    output: 'ok [{"id":1,"name":"TestName2","ip":"0.0.0.1","register_ip":"0.0.0.1","internal_key":"TopSecret","os_name":"TestOsName","os_version":"TestOsVersion","os_major":"TestOsMajor","os_minor":"TestOsMinor","os_codename":"TestOsCodeName","os_build":"TestOsBuild","os_platform":"TestOsPlatfor","os_uname":"TestOsUname","os_arch":"TestOsArch","version":"TestVersion","config_sum":"TestConfigSum","merged_sum":"TestMergedSum","manager_host":"TestManagerHost","node_name":"TestNodeName","date_add":1599223378,"last_keepalive":*,"status":"updated","fim_offset":100,"reg_offset":100,"group":"TestGroup2","sync_status":"syncreq"}]'
```
DProvinciani, line 102: Considering that just before this stage there was an `update-keepalive`. Shouldn't be better to have an `active` state? I know that up to this point the `update-keepalive` command was not modified to also set the `connection_status` to `active`, but this will happen in the near future so, it would be good if we update the test cases thinking on that.

DProvinciani (changes requested):
Apply changes according to the comments.

tests/integration/test_wazuh_db/data/global_messages.yaml
DProvinciani: Please, don't take into account this comment because you don't have a way/mechanism to have the status as `active` in this moment.

pereyra-m: We can do that soon.
Thank you!

tests/integration/test_wazuh_db/data/global_messages.yaml
pereyra-m: That is logic, but currently, the **wdb_global_sync_agent_info_set()** method would ignore the field **connection_status**.

DProvinciani (approved):
Looks good!

PR #916, opened 2020-10-23
Adapt RegistryEventChecker class for cases with mtime disabled

jesusjimsa:
## Description

This pull request adapts the `RegistryEventChecker` class to the cases when `check_mtime` is set to `no`. In this cases, the events for the parent registry key will not show up and it would have failed with the previous implementation.

This also simplifies a bit the class and fixes the registry key default validator and adds the registry value one.


bah07 (approved):
LGTM!

Issue #891, opened 2020-10-14
Registry integration tests: Basic usage 


antoniomanuelfr:
After the development of wazuh/wazuh#5576, new integration tests are going to be needed.
Each file performs a single test. The structure of the directory will be:

|File | Test | Action |Expected result |
|---|---|---|---|
|test_basic_usage_registry_baseline_generation| test_wait_until_baseline| Monitor a registry using scheduled scan | Check if events appear after the baseline|
|test_basic_usage_registry_changes|test_registry_changes| Create, update and delete subkeys/values of a monitored key| Events are generated for keys and values| 
|test_basic_usage_registry_delete_registry|test_delete_registry| Monitor a registry key and delete it| Delete events are generated for each subkey (and it's values) and values|
|test_basic_usage_registry_entries_match_count|test_entries_match_key_count|Monitor a key| Check if FIM entries match the number of monitored keys, subkeys, values|
|test_basic_usage_registry_new_key|test_new_key|Monitor a not existing key| After the key is created, events are generated in the next scheduled scan.|
| test_long_registry_path | test_long_registry_path|Monitor a registry with a long path (>> 260) | Check syscheckd detects registry changes (add, modify, delete) |

jesusjimsa:
Basic usage tests will ned to check that alerts are being sent using all value types:

- REG_SZ
- REG_EXPAND_SZ
- REG_MULTI_SZ
- REG_DWORD
- default

PR #944, opened 2020-11-23
Integration tests for registries basic usages

jotacarma90:
Hello team,

This PR is part of the Integration test epic issue to Fim improvement registries.

Closes https://github.com/wazuh/wazuh-qa/issues/891

# Tests logic
- Check if events appear after the baseline
- Events are generated for keys and values
- Delete events are generated for each subkey (and it's values) and values
- Check if FIM entries match the number of monitored keys, subkeys, values
- After the key is created, events are generated in the next scheduled scan.
- Check syscheckd detects registry changes (add, modify, delete)

# Tests checks

- [x] Proven that tests **fail** when they have to fail


Molter73 (approved):
LGTM!

PR #976, opened 2020-12-17
Added cases to test_scan_day_and_time to check noon and midnight

jotacarma90:
Hello team,
This PR adds two new cases to integration test `test_scan_day_and_time` to check 12 noon and 12 midnight hours, due to a bug where the `__gethour` function misinterpreted these values.
So the test checks that this bug does not occur.

These new cases fail when executed:
image

And after the fix:
image


Molter73:
LGTM!

Molter73 (approved):

PR #995, opened 2020-12-24
Improving GCloud Integration Tests stability 

### LINKED ISSUE #994: Analyze and fix the issues reported in the gcloud integration tests
### Description

As part of the epic https://github.com/wazuh/wazuh-qa/issues/979, we have to goal of stabilizing all the integration tests as well as document and organize the repository.

In the execution of the `gcloud` integration tests, we got some failures that should be investigated to determine whether a fix is required or it was an error in the execution. If there is an error in the test cases, a fix for the error should be implemented as part of this issue.

pereyra-m:
Hello team,

After running all the `gcloud` tests in a local environment, it was found that some of them may need a reliability improvement.
These tests were updated:

- **test_rules.py**: the sample data consumed by the test was generating a different list of rules ID than the one previously defined
- **test_remote_configuration.py**: this _gcloud_ test now has a more general way to find the JSON field 'gcp-pubsub'

The rest work as expected. It would be useful in the future to include them in a Jenkins pipeline.

Closes #994 

# Tests checks

- [x] Proven that tests **fail** when they have to fail

Best regards.

### DIFF
--- a/tests/integration/test_gcloud/test_configuration/test_remote_configuration.py
+++ b/tests/integration/test_gcloud/test_configuration/test_remote_configuration.py
@@ -94,13 +94,9 @@ def get_remote_configuration(component_name, config):
     try:
         if rec_msg_ok.startswith('ok'):
             remote_configuration = json.loads(rec_msg)
-            if host_type == 'server':
-                remote_configuration_gcp = remote_configuration['wmodules'][6]['gcp-pubsub']
-            else:
-                if sys.platform == 'darwin':
-                    remote_configuration_gcp = remote_configuration['wmodules'][3]['gcp-pubsub']
-                else:
-                    remote_configuration_gcp = remote_configuration['wmodules'][5]['gcp-pubsub']
+       	    for	element	in remote_configuration['wmodules']:
+       	       	if 'gcp-pubsub'	in element:
+                    remote_configuration_gcp = element['gcp-pubsub']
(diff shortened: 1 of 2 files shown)


tests/integration/test_gcloud/test_configuration/test_remote_configuration.py
```diff
@@ -94,13 +94,9 @@ def get_remote_configuration(component_name, config):
     try:
         if rec_msg_ok.startswith('ok'):
             remote_configuration = json.loads(rec_msg)
-            if host_type == 'server':
```
DProvinciani, line 97: If this check is no longer necessary, shouldn't we remove the assignment in line 79?

palaciosjeremias (approved):

tests/integration/test_gcloud/test_configuration/test_remote_configuration.py
palaciosjeremias: +1

SpotHound (approved):
LGTM

tests/integration/test_gcloud/test_configuration/test_remote_configuration.py
pereyra-m: Fixed!

DProvinciani (approved):
Looks good!

Issue #1046, opened 2021-02-05
Review vulnerability detector tests: test_window_version_indexing fails 


SpotHound:
Build: https://ci.wazuh.info/job/test_integration_vulnerability_detector/309/

![image](https://user-images.githubusercontent.com/15269938/107055318-75604000-67d1-11eb-9aeb-81ac438865a5.png)

```
test_vulnerability_detector/test_scan_results/test_msu_inventory_msu_feed.py:123: in test_vulnerabilities_report
    f" solves {cve}"
/usr/local/lib/python3.6/site-packages/wazuh_testing-4.0.0-py3.6.egg/wazuh_testing/tools/monitoring.py:189: in start
    error_message=error_message).result()
/usr/local/lib/python3.6/site-packages/wazuh_testing-4.0.0-py3.6.egg/wazuh_testing/tools/monitoring.py:476: in start
    raise TimeoutError()
E   TimeoutError
 -----------------------------Captured stderr setup------------------------------ 
2021/02/05 10:21:45 wazuh-modulesd[19900] debug_op.c:70 at _log(): DEBUG: Logging module auto-initialized
2021/02/05 10:21:45 wazuh-modulesd[19900] wmodules-vuln-detector.c:609 at wm_vuldet_read_provider(): DEBUG: Added msu feed. Interval: 10s | Multi path: '/tmp/test_integration_vulnerability_detector_B309_20210205152053/tests/integration/test_vulnerability_detector/test_scan_results/data/custom_msu.json' | Multi url: 'none' | Update since: 0 | Timeout: 300s
2021/02/05 10:21:45 wazuh-modulesd[19900] wmodules-vuln-detector.c:609 at wm_vuldet_read_provider(): DEBUG: Added nvd feed. Interval: 10s | Multi path: '/tmp/test_integration_vulnerability_detector_B309_20210205152053/tests/integration/test_vulnerability_detector/test_scan_results/data/custom_nvd_feed.json' | Multi url: 'none' | Update since: 2010 | Timeout: 300s
2021/02/05 10:21:45 wazuh-modulesd[19902] debug_op.c:70 at _log(): DEBUG: Logging module auto-initialized
[...]

PR #1047, opened 2021-02-05
Avoid problematic race-condition on VD integration tests for Windows

SpotHound:
|Related issue|
|---|
|Closes https://github.com/wazuh/wazuh-qa/issues/1046|

## Description
This PR fixes a problem with the windows vulnerability detector integration tests related to modulesd alive while the test is mocking data into wazuh database.

Modulesd needs to be stoped while mocking the database. Also, we've added a catch for an exception that may occur on the first test of this check due to vulnerability detector not having created the required tables in the database yet. If this occurs, the catch statement will restart modulesd (to force a load of vulnerability detector if it wasn't already loaded) and wait a few seconds for the tables to be created. 

## Tests

image

https://ci.wazuh.info/job/test_integration_vulnerability_detector/310/console


jmv74211 (approved):
Good job!

The updated fixture 'template' is used in other scan tests too. It may be that these changes can solve the problems of false positives that sometimes occur randomly.

We have to test with the rest and see the results.

BraulioV (approved):

PR #1084, opened 2021-02-24
Fix some problems with Agent simulator

### LINKED ISSUE #1082: Agent simulator bug
Agent simulator received messages from the manager and processes them. If a message has one reserved word (for example, `upgrade` or `com`) it will be considered as a command and will be processed.

This is a problem for two reasons:

1. It may be that the agent receives entire files from the manager and they could contain these words (`upgrade` is a very common word).
2. Sometimes when the manager send long messages (for example, an entire file) they will be split into different messages and this could lead to words such as "compromised" being cut into "com" and "promised", an this would make the first message to be processed because it contains `com`.

We need to fix it soon.

SpotHound:
|Related issue|
|---|
|Closes #1082|

## Description

This PR fixes some problems related to the agent simulator.

## Tests

- [x] tests/integration/test_rids/test_rids.py

- [x] tests/integration/test_rootcheck/test_rootcheck.py

- [x] tests/integration/test_wpk/test_wpk_manager_task_states.py

- [x] tests/integration/test_wpk/test_wpk_manager.py

## Notes
During the development of these changes, we've faced some issues related to python and memory allocation. The kernel of the running machine (a centos7 virtual machine for testing) was killing the pytest process due to large amounts of data being allocated. We've made some tests with the master branch and we've concluded that this problem isn't related to my changes but is caused by a bad memory administration on the current agent_simulator code. I will open another issue to investigate this.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
@@ -28,7 +28,7 @@ from time import mktime, localtime, sleep, time
 from wazuh_testing.tools.remoted_sim import Cipher
 
 logging.basicConfig(
-    level=logging.INFO,
+    level=logging.info,
     format="%(asctime)s:%(levelname)s:AGENT_SIMULATOR:%(message)s"
 )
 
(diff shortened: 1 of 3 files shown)


deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
```diff
@@ -199,7 +209,7 @@ def register(self):
         self.key = registration_info[3]
         ssl_socket.close()
         sock.close()
-        print("Registration - {}({})".format(self.name, self.id))
+        logging.info("Registration - {}({})".format(self.name, self.id))
```
jmv74211, line 212: I would use `logging.debug()` for all debug messages, because all these messages will be printed even if debug of the agent simulator is set to false

BraulioV (approved):

PR #1086, opened 2021-02-24
Add integration tests for syslog

### LINKED ISSUE #1083: Remoted tier 0: tests for syslog in remoted
`wazuh-remoted` can configure a Syslog server to receive messages from external components and generate alerts for your infrastructure without installing an agent.

This test must check these two points:
- [x] remoted can receive events using UDP.
- [x] remoted can receive events using TCP.

BraulioV:
|Related issue|
|---|
|closes #1083 |

## Description
This PR adds new integration tests for the `syslog` option of remoted. These checks test if `remoted` is able to receive messages to its Syslog server and log them in the `archives.log`.

## Configuration options

NA

## Logs example
image

## Tests

- [x] The test is documented in wazuh-qa/docs

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/remote.py
+++ b/deps/wazuh_testing/wazuh_testing/remote.py
@@ -1,0 +1,102 @@
+# Copyright (C) 2015-2021, Wazuh Inc.
+# Created by Wazuh, Inc. <info@wazuh.com>.
+# This program is free software; you can redistribute it and/or modify it under the terms of GPLv2
+
+import socket
+
+from wazuh_testing.tools import ARCHIVES_LOG_FILE_PATH
+from wazuh_testing.tools.file import truncate_file
+from wazuh_testing.tools.monitoring import FileMonitor, make_callback, REMOTED_DETECTOR_PREFIX
+
+
+def callback_detect_remoted_started(port, protocol, connection_type="secure"):
+    """Creates a callback to detect if remoted was correctly started
+
+    wazuh-remoted logs if it has correctly started for each connection type, the port and
+    the protocol in the ossec.log
+
+    Args:
+        port (int): port configured for wazuh-remoted.
+        protocol (str): protocol configured for wazuh-remoted. It can be UDP, TCP or both options at the same time.
+        connection_type (str): it can be secure or syslog.
+
+    Returns:
+        callable: callback to detect this event
+    """
+    msg = fr"Started \(pid: \d+\). Listening on port {port}\/{protocol} \({connection_type}\)."
+
+    return make_callback(pattern=msg, prefix=REMOTED_DETECTOR_PREFIX)
+
+
+def callback_detect_syslog_event(message):
+    """Creates a callback to detect the syslog messages in the archives.log
+
+    Args:
+        message (str): syslog message sent through the socket
+
+    Returns:
+        callable: callback to detect this event
+    """
+    expr = fr".*->\d+\.\d+\.\d+\.\d+\s{message}"
+    return make_callback(pattern=expr, prefix=None)
+
+
+def send_syslog_message(message, port, protocol, manager_address="127.0.0.1"):
+    """This function sends the ping message to the manager
+
+    This message is the first of many between the manager and the agents. It is used to check if both of them are ready
+    to send and receive other messages
+
+    Args:
+        message (str): string to send as a syslog event.
+        protocol (str): it can be UDP or TCP.
+        port (int): port where the manager has bound the remoted port
+        manager_address (str): address of the manager.
+
+    Raises:
+        ConnectionRefusedError: if there's a problem while sending messages to the manager
+    """
+    if protocol == "UDP":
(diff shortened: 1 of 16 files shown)


deps/wazuh_testing/wazuh_testing/remote.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Returns:
+        callable: callback to detect this event
+    """
+    expr = fr".*->\d+\.\d+\.\d+\.\d+\s{message}"
+    return make_callback(pattern=expr, prefix=None)
+
+
+def send_syslog_message(message, port, protocol, manager_address="127.0.0.1"):
+    """This function sends the ping message to the manager
```
SpotHound, line 45: Change this, I think that was copy-pasted from `send_ping_pong_messages`

SpotHound (approved):
lgtm, just a little suggestion.

deps/wazuh_testing/wazuh_testing/remote.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    # Reset ossec.log and start a new monitor
+    truncate_file(ARCHIVES_LOG_FILE_PATH)
+    wazuh_archives_log_monitor = FileMonitor(ARCHIVES_LOG_FILE_PATH)
+
+    return wazuh_archives_log_monitor
+
+
+def detect_archives_log_event(archives_monitor, callback, error_message, update_position=True, timeout=5):
+    archives_monitor.start(timeout=timeout, update_position=update_position, callback=callback,
```
jmv74211, line 95: Missing documentation?

SpotHound (approved):
lgtm

deps/wazuh_testing/wazuh_testing/remote.py
BraulioV: Touché. Done in 3f381f0add7acadf3c21a21d945322ac62541e4e

deps/wazuh_testing/wazuh_testing/remote.py
```diff
@@ -0,0 +1,136 @@
...
+    Args:
+        message (str): string to send as a syslog event.
+        protocol (str): it can be UDP or TCP.
+        port (int): port where the manager has bound the remoted port
+        manager_address (str): address of the manager.
+
+    Raises:
+        ConnectionRefusedError: if there's a problem while sending messages to the manager
+    """
+    if protocol == "UDP":
```
jmv74211, line 56: If I enter 'udp' string, then it would go into the else part, which would not be correct. We have to think about defining a standard or always use a string upper for these cases

deps/wazuh_testing/wazuh_testing/remote.py
```diff
@@ -0,0 +1,136 @@
...
+    return wazuh_archives_log_monitor
+
+
+def detect_archives_log_event(archives_monitor, callback, error_message, update_position=True, timeout=5):
+    archives_monitor.start(timeout=timeout, update_position=update_position, callback=callback,
+                           error_message=error_message)
+
+
+def check_syslog_event(wazuh_archives_log_monitor, message, port, protocol):
+    """Check if a syslog event is properly receive by the manager.
```
jmv74211, line 100:
```suggestion
    """Check if a syslog event is properly received by the manager.
```

deps/wazuh_testing/wazuh_testing/remote.py
```diff
@@ -0,0 +1,136 @@
...
+        wazuh_archives_log_monitor (FileMonitor): FileMonitor object to monitor the archives.log.
+        message (str): Message sent for syslog that must appear in the archives.log.
+        protocol (str): it can be UDP or TCP.
+        port (int): port where the manager has bound the remoted port
+    """
+    send_syslog_message(message, port, protocol)
+    detect_archives_log_event(archives_monitor=wazuh_archives_log_monitor,
+                              callback=callback_detect_syslog_event(message),
+                              error_message="Syslog message wasn't received or took too much time.")
+
```
jmv74211, line 112: The function names may cause some confusion. For now, all events have been searched for in the ossec.log, and particularly this function searches for them in the archives.log. We have to think of something to know with the name where they are being looked for. In this case something like `check_syslog_event_in_archives`

BraulioV: This function can be used to search for any event in the `archives.log` where many of the events received by remoted are logged. That's why there's no point to change the name to `check_syslog_event_in_archives` when it can be used to search any event. 

In the future, we should add a new module to `wazuh_testing` for searching events in the log that can be used by any other module, however, this is no the time to do it.

deps/wazuh_testing/wazuh_testing/remote.py
BraulioV: This has already been done at https://github.com/wazuh/wazuh-qa/commit/3f381f0add7acadf3c21a21d945322ac62541e4e

SpotHound (approved):

Issue #1062, opened 2021-02-15
Remoted tier 0: manager -> agent communication


BraulioV:
These tests will check if remoted can properly send messages to the agent and collect the agent's response. 

To do so, we will use the socket used by the API to request information about the agent. To give more info about this, when the user runs an API request that needs to communicate or collect info from the agent, the API writes its request in this socket `$DIR/queue/ossec/request` and then, remoted forwards this query to the agent using the `secure` socket bound to the port 1514. 

Finally, the agent sends its response to remoted and it forwards it to the API. Here you can see the messages flow:
![image.png](https://images.zenhubusercontent.com/5d91c1bf3ab6b200010aecd1/d3ab1079-6d5a-48b9-8dd9-a9259e0961d4)

With this, the test must perform these checks:
- [x] Write a request in `$DIR/queue/ossec/request` and check if remoted forwards it to the agent, collects the response, and writes it in the socket.
- [x] Write a request in `$DIR/queue/ossec/request` to collect statistics from the agent, check if remoted forwards it to the agent, collects the response, and writes it in the socket.
- [x] Write a request in `$DIR/queue/ossec/request` for a disconnected agent. remoted must respond with a message showing that the agent is not connected.

SpotHound:
Hello team, 

I'm working on this at `1062-agent-communication`.

I'm stating with the disconnected agent test as it should be easier to implement.

SpotHound:
Ping example

```
msg = '#ping'
msg_size = len(bytearray(msg, 'utf-8'))
ping_msg = msg_size.to_bytes(4, 'little') + msg.encode()
s.send(ping_msg)
response = s.recv(len(ping_msg))
```

SpotHound:
### pip freeze 
```
asn1crypto==0.24.0
attrs==20.3.0
backcall==0.2.0
certifi==2020.12.5
cffi==1.14.5
chardet==3.0.4
configobj==5.0.6
cryptography==3.4.6
dbus-python==1.2.4
decorator==4.2.1
filetype==1.0.7
gpg==1.10.0
idna==2.5
importlib-metadata==3.4.0
iniconfig==1.1.1
ipython==7.16.1
ipython-genutils==0.2.0
isc==2.0
jedi==0.18.0
lockfile==0.12.2
netifaces==0.10.6
packaging==20.9
parso==0.8.1
pciutils==2.3.6
perf==0.1
pexpect==4.8.0
pickleshare==0.7.5
[...]

PR #1088, opened 2021-02-26
Remoted tier 0: manager -> agent communication

SpotHound:
|Related issue|
|---|
|Closes https://github.com/wazuh/wazuh-qa/issues/1062|

## Description

This PR adds some checks for manager-agent communication through remoted for tier 0 of remoted integration tests. It also:

- Fix an issue with `get_version` method on tools
- Adds logging methods for some modules (including agent simulator)
- Fix syntax errors on wpk tests
- Add new methods to wazuh_db module

All the tests are working now
image

Just two problems to solve in the future:
- The test need the Wazuh manager to consider mocked agent as active but, we cannot mock them into the global db without stopping wazuh_db and modulesd, so we need to wait for the manager to add them after the registration (this can take up to 1 min) this makes the test very slow and could be improved
- Sometimes remoted answer with an empty message in the "disconnect agent" tests when the query is made too early. We've solved this problem by adding some retries to the query if the answer is empty. 

## Configuration options

```
parameters = [
    {'PROTOCOL': 'udp,tcp'},
    {'PROTOCOL': 'tcp'},
    {'PROTOCOL': 'udp'},
]

metadata = [
    {'PROTOCOL': 'udp,tcp'},
    {'PROTOCOL': 'tcp'},
    {'PROTOCOL': 'udp'},
]

```
#### Api call
We make the following API call
```
TOKEN=$(curl -u wazuh:wazuh -k -X GET "https://localhost:55000/security/user/authenticate?raw=true")
curl -k -X GET "https://localhost:55000/agents/001/config/agent/client" -H "Authorization: Bearer $TOKEN"
```

#### Remoted request

API writes to `/var/ossec/queue/ossec/requests` and remoted reads:

```
b'\x1a\x00\x00\x00001 agent getconfig client'
```

#### Agent config request

Remoted send to agent the same message, only header and format changes to:

```
ececac937b8e5dead15e9096e8bd5215214970000000002:3090:#!-req c2b2c9e3 agent getconfig client
```

#### Agent response

Agent answer to remoted with:

```
#!-req 12345 ok {"internal":{"agent":{"debug":0,"warn_level":90,"normal_level":70,"tolerance":15,"recv_timeout":60,"state_interval":5,"min_eps":50,"remote_conf":1},"monitord":{"rotate_log":1,"compress":1,"keep_log_days":31,"day_wait":10,"size_rotate":512,"daily_rotations":12},"remoted":{"request_pool":1024,"request_rto_sec":1,"request_rto_msec":0,"max_attempts":4,"comp_average_printout":19999,"recv_counter_flush":128,"verify_msg_id":0}}}
```

#### API result

And finally, remoted send the relevant information to the API:

```
Out[25]: b'\xbb\x01\x00\x00ok {"client":{"config-profile":"centos, centos8, centos8.3","notify_time":10,"time-reconnect":60'
```

## Tests

- [x] The test is documented in wazuh-qa/docs


deps/wazuh_testing/wazuh_testing/tools/sockets.py
```diff
@@ -0,0 +1,23 @@
...
+
+def send_request(msg_request, response_size=100):
+    sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    request_msg = struct.pack('<I', len(msg_request)) + msg_request.encode()
+
+    sock.connect(request_socket)
+    sock.send(request_msg)
+    response = sock.recv(response_size).decode()
+    sock.close()
+
```
BraulioV, line 21: It would be interesting to compute the message size using the header and then receive the exact size.

BraulioV (approved):

BraulioV (approved):

BraulioV (approved):

PR #1092, opened 2021-02-26
Adding new tests for Wazuh-DB insert commands in agents' CVEs table

palaciosjeremias:
|Related issue|
|---|
|1069|

## Description

This PR adds test cases for the following vuln_cve insert commands:
- Insert command with a paylod that contains white space (to validate a fix in the parsing of vuln_cve commands)
- Insert command without payload.

## Tests


DProvinciani (approved):
Looks good!

PR #1159, opened 2021-03-18
Review of rootcheck module on agent simulator

pacoqack:
|Related issue|
|---|
|Closes 1157|

Just some minor changes on the rootcheck module on the agent simulator to make sure that it is able to send the expected amount of event (enough to generate the desired EPS rate or received events for some seconds). IF the rootcheck frequency is set to `0` it will send messages without stoping at a fixed EPS rate. 

We've made the changes so it is possible to define a stressed scenario where the interval value still has an effect. For example, if the test duration is 5 minutes and the interval parameter is set to 60 seconds and the selected EPS is 100, the simulated agent will send 500 events in the first 36 seconds and then wait for the remaining seconds (0EPS).

The value for stressing time is set to a 0.5 of the selected frequency but could be redefined in the future if needed.

In this scenario the EPS evolution on time would be something like this graph:

Figure_1

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
@@ -71,10 +71,10 @@ class Agent:
         short_version (str): Agent version in format x.y
         cypher (str): Encryption method for message communication.
         os (str): Agent operating system.
-        fim_eps (int): Set the fim's maximum event reporting throughput.
-        fim_integrity_eps (int): Set the fim_integrity's maximum event reporting throughput.
-        syscollector_eps (int): Set the syscollector's maximum event reporting throughput.
-        rootcheck_eps (int): Set the rootcheck's maximum event reporting throughput.
+        fim_eps (int): Fim's maximum event reporting throughput. Default `1000`.
+        fim_integrity_eps (int): Fim integrity's maximum event reporting throughput. Default `100`.
+        syscollector_eps (int): Syscollector's maximum event reporting throughput. Default `100`.
+        rootcheck_eps (float): Rootcheck's maximum event reporting throughput. Default `60.0`.
         manager_address (str): Manager IP address.
         encryption_key (bytes): Encryption key used for encrypt and decrypt the message.
         keep_alive_event (bytes): Keep alive event (read from template data according to OS and parsed to an event).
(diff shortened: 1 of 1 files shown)


deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
```diff
@@ -71,8 +71,10 @@ class Agent:
         short_version (str): Agent version in format x.y
         cypher (str): Encryption method for message communication.
         os (str): Agent operating system.
-        fim_eps (int): Set the maximum event reporting throughput. Events are messages that will produce an alert.
-        fim_integrity_eps (int): Set the maximum database synchronization message throughput.
+        fim_eps (int): Set the fim's maximum event reporting throughput.
+        fim_integrity_eps (int): Set the fim_integrity's maximum event reporting throughput.
+        syscollector_eps (int): Set the syscollector's maximum event reporting throughput.
+        rootcheck_eps (int): Set the rootcheck's maximum event reporting throughput.
```
Rebits, line 77: Add default values to the documentation

```suggestion
        fim_eps (int): Fim's maximum event reporting throughput. Default `1000`.
        fim_integrity_eps (int): Fim integrity's maximum event reporting throughput. Default `100`.
        syscollector_eps (int): Syscollector's maximum event reporting throughput. Default `100`.
        rootcheck_eps (float): Rootcheck's maximum event reporting throughput. Default `60.0`.
```

deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
```diff
@@ -133,7 +137,7 @@ def __init__(self, manager_address, cypher="aes", os=None, inventory_sample=None
             "fim": {"status": "enabled", "eps": self.fim_eps},
             "fim_integrity": {"status": "disabled", "eps": self.fim_integrity_eps},
             "syscollector": {"status": "disabled", "frequency": 60.0, "eps": self.syscollector_eps},
-            "rootcheck": {"status": "disabled", "frequency": 60.0, "eps": self.rootcheck_eps},
+            "rootcheck": {"status": "disabled", "frequency": rootcheck_frequency, "eps": self.rootcheck_eps},
```
Rebits, line 140: Follow the used format for the rest of the Agent attributes

```suggestion
            "rootcheck": {"status": "disabled", "frequency": self.rootcheck_frequency, "eps": self.rootcheck_eps},
```

Rebits (commented):
GJ!, but some changes are required

Rebits (approved):

pacoqack:
ty @Rebits

PR #1196, opened 2021-04-05
Fixed size for all agent simulator module messages

### LINKED ISSUE #1175: Agent simulator: Fixed length messages
It is necessary for the implementation of performance tests a new `agent_simulator` attribute. This attribute will fix the size of the agent's messages. This size should be used for all types of events that the agent's simulator sends to the manager. 

For the implementation of this new attribute the attribute `labels` will be used, in such a way that as many labels as necessary will be added so that the message has the desired size

### Update

Due to labels are only sent by the agent in `keep-alive` messages, agent simulator will implement fixed-size messages using extra random characters appended to the messages

Rebits:
|Related issue|
|---|
|closes #1175    |

This PR adds a `fixed_message_size` attribute to the agent simulator. This allows set fixed size in KB for all agent messages. It is implemented by adding a string of character 'A'  at the end of the message.

In addition, the `simulate_agent` script has been modified to include this new parameter


deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py
```diff
@@ -175,6 +179,7 @@ def __init__(self, manager_address, cypher="aes", os=None, rootcheck_sample=None
         self.stage_disconnect = None
         self.setup(disable_all_modules=disable_all_modules)
         self.rcv_msg_queue = Queue(rcv_msg_limit)
+        self.fixed_message_size = fixed_message_size
```
BraulioV, line 182: This value can be stored in bytes to avoid computing it in the loop.

BraulioV (changes requested):

spothound (approved):

BraulioV (approved):

PR #1214, opened 2021-04-12
Add API performance tests

vicferpoy:
|Related issue|
|---|
| https://github.com/wazuh/wazuh/issues/7940 |

## Description

Hello team, 

We added a new set of tests for performance. As for the moment, we only have the API tests. These will test a bunch of Wazuh API endpoints on another machine to obtain certain metrics and results to see which endpoints are scalable and make the necessary improvements.

## Configuration options

These tests are configurable through the `data/configuration.yaml` file. This includes the following options:

```yaml
configuration:
  host: 'localhost'
  port: 55000
  restart_delay: 30

test_cases:
  - endpoint: /active-response
    method: put
    parameters: {}
    body:
      command: custom
      custom: True
    restart: False

  - endpoint: /agents
    method: get
    parameters: {}
    body: {}
```

More information about this file can be found in the README.

## Report example

These tests are designed to use the `--html=report.html --self-contained-html` options, as we want to see certain metadata that is only accessible through this. This would be an example of the report:

image

## Tests

- [ ] The test is documented in wazuh-qa/docs.
- [ ] `provision_documentation.sh` generate the docs without errors.

Regards,
Víctor Fernández

### DIFF
--- a/tests/performance/test_api/README.md
+++ b/tests/performance/test_api/README.md
@@ -1,0 +1,74 @@
+# wazuh-qa
+
+Wazuh - System quality assurance automation templates
+
+## Setting up a test environment
+
+To run these tests we need to use a **Linux** machine with direct connection to another machine with Wazuh installed.
+
+### Dependencies
+
+In addition, we need the `wazuh-testing` package in addition to the default requirements from this repository.
+
+## Performance API tests
+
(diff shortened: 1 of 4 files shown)


davidjiglesias (changes requested):
Add configurable token duration

davidjiglesias (approved):

BraulioV (approved):

PR #1239, opened 2021-04-19
Send syscollector events with a constant throughput 

### LINKED ISSUE #1238: Performance tests: Syscollector events must be sent with a constant throughput
The current implementation of the `simulate-agents` tool generates events for Syscollector using batches. These batches will send events with a fixed EPS during a specific amount of time. By default, it sends x _EPS_ for one minute. After that, it sleeps for a while before sending a new batch.

image
image

This is ok but we should change this behavior to a constant throughput to keep consistency between all the different modules.

BraulioV:
|Related issue|
|---|
|closes #1238|

## Description

This PR closes #1238 by setting the Syscollector's frequency in the `simulate-agents` tool to 0. These lines https://github.com/wazuh/wazuh-qa/blob/2f73335e4bf5a6008481a6401ef3a6689055303d/deps/wazuh_testing/wazuh_testing/tools/agent_simulator.py#L1606-L1609 control how the events are sent and the batch size. By setting the frequency to 0 we can send a constant throughput of events.

## Configuration options

NA
## Logs example

image

## Tests

- [x] The test is documented in wazuh-qa/docs.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/scripts/simulate_agents.py
+++ b/deps/wazuh_testing/wazuh_testing/scripts/simulate_agents.py
@@ -35,7 +35,7 @@ def run_agents(agents_number=1, manager_address='localhost', protocol=TCP, agent
 
     for _ in range(agents_number):
         agent = ag.Agent(manager_address, "aes", os=agent_os, version=agent_version, fim_eps=eps,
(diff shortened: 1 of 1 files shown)


Rebits (approved):
LGTM

Issue #1281, opened 2021-05-03
 Integrate the `simulate_api` tool with the cluster performance test pipeline


jmv74211:
It is requested to integrate the `simulate_api` tool with the cluster performance pipeline. This is necessary to stress the API while the API performance tests are being launched.

To do this, the following tasks are required:
- [x] **1**: Add new config parameters to the cluster performance test pipeline UI.
- [x] **2**: Create a Dockerfile, entry-point script...
- [x] **3**: Add a new parameter to specify the running time for the simulate-api script.
- [x] **4**: Create and upload the docker to ECR.
- [x] **5**: Create a new task definition for launching the docker images as ECS Tasks.
- [x] **6**: Add the new ECSTasks to the cluster performance tests pipeline.
- [x] **7**: Add custom log path in API load simulator tool and script.
- [x] **8**: Add API load simulator logs to the Jenkins artifacts.
- [x] **9**: Testing and debugging.

PR #1293, opened 2021-05-04
Remove timeout parameter and add a new one to set the log path in API load simulator tool and script

jmv74211:
|Related issue|
|---|
|#1281|

This PR makes the following changes:
- Remove the `timeout` parameter from `simulate_api_load` and `api_simulator`. The main reason is to wait for the `timeout` from the `requests` library.

- Add a new parameter to set the custom log path for the simulation info (Necessary to store these logs as Jenkins artifacts in the cluster performance test).


vicferpoy (approved):
LGTM!

BraulioV (approved):

PR #1331, opened 2021-05-13
[FrontEnd] Added functionality to use different login methods and refactored some functions

MauGaP:
|Related issue|
|---|
|---|

## Description
Added a switch to use different login methods. and refactored the functions to be more reusable

## Configuration options
Note that in the cypress.json you can modify the url in which you search for your application.


tests/frontend/test_kibana_app/package.json
```diff
@@ -11,12 +11,16 @@
...
   },
   "devDependencies": {
-    "cypress-cucumber-preprocessor": "^4.0.3"
+    "cypress-cucumber-preprocessor": "^4.0.3",
+    "eslint-plugin-cypress": "^2.11.3"
   },
   "cypress-cucumber-preprocessor": {
-    "nonGlobalStepDefinitions": true
+    "nonGlobalStepDefinitions": false,
+    "step_definitions": "cypress/integration/step_definitions/**/"
```
matiasmoreno876, line 24: Perfect approach

tests/frontend/test_kibana_app/cypress/integration/utils/constants.js
```diff
@@ -0,0 +1,7 @@
+export const BASIC = 'basic';
+export const ODFE = 'odfe';
+export const ODFE_PASSWORD = 'admin';
+export const ODFE_USERNAME = 'admin';
+export const XPACK = 'xpack';
+export const XPACK_PASSWORD = 'elastic';
+export const XPACK_USERNAME = 'elastic';
```
matiasmoreno876, line 7: `User` and `password` should be obtained from .env files

MauGaP: This is a good idea. But we will continue to use the current approach because nowadays our environments are generated from a docker-compose file that has defined these constants already defined. 
A better approach would be to read those files and get the strings from there.
This will be useful when moving to a CI env.

tests/frontend/test_kibana_app/package.json
MauGaP: thanks!

matiasmoreno876 (approved):
LGTM

PR #1345, opened 2021-05-17
[Fix] control_service doesn't manage Wazuh daemons correctly

### LINKED ISSUE #1316: Auth integration tests error
image

There is an error in our auth integration tests when running on the master branch.

The problem is that the `configure_sockets_environment` function calls to `check_daemon_status` after stopping wazuh and this function has a timeout (by default) of 10 seconds. As some daemons (in this case, clusters) have not stopped in these 10 seconds. The test fails before even start.

We need to increase the timeout of `check_daemon_status` function and check fi this sovles the problem.

mdengra:
|Related issue|
|---|
|Closes #1316|

## Description

This PR fixes the problem with the `control_service` function in which it is not able to handle Wazuh daemons running through Python, causing some tests, such as `test_authd` to fail when trying to terminate the `wazuh-clusterd` daemon, which runs through a Python module and not as a standalone process.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/monitoring.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/monitoring.py
@@ -779,8 +779,8 @@ class ManInTheMiddle:
 
         # Give proper permissions to socket
         if isinstance(self.listener_socket_address, str):
-            uid = pwd.getpwnam('ossec').pw_uid
-            gid = grp.getgrnam('ossec').gr_gid
+            uid = pwd.getpwnam('wazuh').pw_uid
+            gid = grp.getgrnam('wazuh').gr_gid
             os.chown(self.listener_socket_address, uid, gid)
             os.chmod(self.listener_socket_address, 0o660)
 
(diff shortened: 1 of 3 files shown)


spothound (approved):
Awesome work @mdengra! Thank you very much.

spothound:
## Jenkins testing

Example of failed checks: https://ci.wazuh.info/job/Test_integration/3796/
Replay with the fix: https://ci.wazuh.info/job/Test_integration/3874/console

deps/wazuh_testing/wazuh_testing/tools/monitoring.py
```diff
@@ -779,8 +779,8 @@ def run(self, *args):
 
         # Give proper permissions to socket
         if isinstance(self.listener_socket_address, str):
-            uid = pwd.getpwnam('wazuh').pw_uid
-            gid = grp.getgrnam('wazuh').gr_gid
+            uid = pwd.getpwnam('ossec').pw_uid
+            gid = grp.getgrnam('ossec').gr_gid
```
BraulioV, line 783: Revert this change

BraulioV (changes requested):

deps/wazuh_testing/wazuh_testing/tools/monitoring.py
mdengra: Done in: https://github.com/wazuh/wazuh-qa/pull/1345/commits/ba4f4aa7304d2cf27ce9776b834e87e1cbeed8c4

BraulioV (approved):

PR #1517, opened 2021-06-28
Fix FIM registry sync IT

### LINKED ISSUE #1343: FIM IT: test registry responses failing
Hello team!

The test `test_registry_responses_win32` is failing due to an error in the test logic: 

The message `dbsync no data` is not sent by the manager if an alert is triggered, as the DB is updated. This provokes the failure of the tests (they passed before because the FIM behavior wasn't correct). 

As the DB can't be updated while FIM is executing (FIM never closes the DB while it's running) the only way to check the behavior is to stop the windows agent, change the monitored keys and start the agent again (this is what the second case does).

We should remove the first cases of this test and check any other failure in the second case.

antoniomanuelfr:
|Related issue|
|---|
|#1343|

## Description
Hi team.

This PR aims to fix the test `test_registry_responses_win32.py` that was constantly failing.
Closes #1343 

## Tests

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/fim.py
+++ b/deps/wazuh_testing/wazuh_testing/fim.py
@@ -1036,15 +1036,15 @@ def callback_detect_integrity_event(line):
 
 
 def callback_detect_registry_integrity_state_event(line):
-    match = re.match(r'.*Sending integrity control message: {"component":"fim_registry","type":"state","data":(.+)}', line)
-    if match:
-        return json.loads(match.group(1))
+    event = callback_detect_integrity_event(line)
+    if event and event['component'] == 'fim_registry' and event['type'] == 'state':
+        return event['data']
     return None
 
 
 def callback_detect_registry_integrity_clear_event(line):
-    match = re.match(r'.*Sending integrity control message: {"component":"fim_registry","type":"integrity_clear".*', line)
-    if match:
+    event = callback_detect_integrity_event(line)
+    if event and event['component'] == 'fim_registry' and event['type'] == 'integrity_clear':
         return True
     return None
 
--- a/deps/wazuh_testing/wazuh_testing/fim.py
+++ b/deps/wazuh_testing/wazuh_testing/fim.py
@@ -1035,10 +1035,17 @@ def callback_detect_integrity_event(line):
     return None
 
 
-def callback_detect_registry_integrity_event(line):
+def callback_detect_registry_integrity_state_event(line):
     match = re.match(r'.*Sending integrity control message: {"component":"fim_registry","type":"state","data":(.+)}', line)
     if match:
         return json.loads(match.group(1))
+    return None
+
+
+def callback_detect_registry_integrity_clear_event(line):
+    match = re.match(r'.*Sending integrity control message: {"component":"fim_registry","type":"integrity_clear".*', line)
+    if match:
+        return True
     return None
 
 
--- a/tests/integration/test_fim/test_synchronization/test_registry_responses_win32.py
+++ b/tests/integration/test_fim/test_synchronization/test_registry_responses_win32.py
@@ -17,14 +17,14 @@ pytestmark = [pytest.mark.win32, pytest.mark.tier(level=1)]
 
 # variables
 key = "HKEY_LOCAL_MACHINE"
-subkey = "SOFTWARE\\random_key"
+monitored_key = "SOFTWARE\\random_key"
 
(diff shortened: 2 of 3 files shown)


deps/wazuh_testing/wazuh_testing/fim.py
```diff
@@ -1035,13 +1035,20 @@ def callback_detect_integrity_event(line):
...
     if match:
         return json.loads(match.group(1))
     return None
 
 
+def callback_detect_registry_integrity_clear_event(line):
+    match = re.match(r'.*Sending integrity control message: {"component":"fim_registry","type":"integrity_clear".*', line)
+    if match:
+        return True
+    return None
```
Molter73, line 1049: We can change this function and the one above it to use `callback_detect_integrity_event` to get the JSON out of the message and use that to make any additional verifications we want, it would look something like this:
```python
def callback_detect_registry_integrity_state_event(line):
    event = callback_detect_integrity_event(line)
    if event and event['component'] == 'fim_registry' and event['type'] == 'state':
        return event['data']
    return None


def callback_detect_registry_integrity_clear_event(line):
    event = callback_detect_integrity_event(line)
    if event and event['component'] == 'fim_registry' and event['type'] == 'integrity_clear':
        return True
    return None
```

tests/integration/test_fim/test_synchronization/test_registry_responses_win32.py
```diff
@@ -187,20 +141,22 @@ def test_registry_sync_after_restart(key_name, value_name, tags_to_apply,
...
     """
     check_apply_test(tags_to_apply, get_configuration['tags'])
-
     key_path = os.path.join(subkey, key_name)
     value_path = os.path.join(key, key_path, value_name)
 
     # wait until the sync is done.
-    get_sync_msgs(sync_interval)
+    wazuh_log_monitor.start(timeout=sync_interval + 15, callback=fim.callback_detect_registry_integrity_clear_event,
+                            error_message='Did not receive expected "integrity clear" event')
```
Molter73, line 149: We should probably move this validation to the fixture in charge of removing the created registry entry.

Molter73 (changes requested):

tests/integration/test_fim/test_synchronization/test_registry_responses_win32.py
```diff
@@ -187,20 +150,18 @@ def test_registry_sync_after_restart(key_name, value_name, tags_to_apply,
...
     key_path = os.path.join(subkey, key_name)
     value_path = os.path.join(key, key_path, value_name)
 
-    # wait until the sync is done.
-    get_sync_msgs(sync_interval)
     # stops syscheckd
     control_service('stop')
-
+    fim.create_registry(fim.registry_parser[key], subkey, fim.KEY_WOW64_64KEY)
     key_handle = fim.create_registry(fim.registry_parser[key], key_path, fim.KEY_WOW64_64KEY)
```
Molter73, line 159: This is a little confusing, I understand that the configured key must not be created prior to starting the test in order for the agent to report an empty database. Maybe we can rename `subkey` to `monitored_key`?

tests/integration/test_fim/test_synchronization/test_registry_responses_win32.py
```diff
@@ -102,75 +103,37 @@ def find_value_in_event_list(key_path, value_name, event_list):
...
+    fim.delete_registry(fim.registry_parser[key], subkey, fim.KEY_WOW64_64KEY)
+    control_service('stop')
+    truncate_file(fim.LOG_FILE_PATH)
+    file_monitor = FileMonitor(fim.LOG_FILE_PATH)
+    setattr(request.module, 'wazuh_log_monitor', file_monitor)
+    control_service('start')
 
-        events = get_sync_msgs(sync_interval + 15)
+    # wait until the sync is done.
+    wazuh_log_monitor.start(timeout=sync_interval + 15, callback=fim.callback_detect_registry_integrity_clear_event,
```
Molter73, line 126: I'm not really fond of this hardcoded `+ 15`, can we replace this with something like the default_timeout?

Molter73 (changes requested):

Molter73 (approved):
LGTM!

Rebits:
GJ, a new [PR](https://github.com/wazuh/wazuh-qa/pull/1663) has been created in order to fix this issue for the `4.2` version. In the future, these changes will be applied to the `master` branch.

Issue #1608, opened 2021-07-16
Create a Config for the DocGenerator module


palaciosjeremias:
DocGenerator must contain a **config.yaml** file and a **config.py** module to parse the values.
The config file must have the below format:

```
Project path: "../../tests/integration"

Documentation path: ".."

Include paths:
  - wazuh_db:
    path: "../../tests/integration/test_wazuh_db"
    recursive: false
  - vulnerability_detector:
    path: "../../tests/integration/test_vulnerability_detector"

Include regex:
  - "^test_.*py$|readme.md$"

Ignore paths:
  - "../../tests/integration/test_wazuh_db/data"
  - "*/data/*"

Valid tags:
  - DB
  - Feeds
  - VulDet
  - Vulnerability Detector

Required fields:
  Module:
    - Brief
    - Metadata
    - Metadata.Modules
    - Metadata.Daemons
    - Metadata.Operating System
    - Metadata.Tiers
  Tests:
    - Test Logic
    - Checks

Ignored fields:
  Case Sensitive: false
  Module:
    - Copyright
  Tests:
    - Params
    - Args
```

Issue #1616, opened 2021-07-19
Implement a sanity and coverage module


MiguelazoDS:
As part of the epic [Overall improvements for integration tests documentation](https://github.com/wazuh/wazuh-qa/issues/1536) we need to implement the module that is in charge of checking the coverage and sanity of the documentation.

The requirements are:

- The coverage function must evaluate how much of the already implemented tests are explained in the documentation. 
  - It should display comprehensive details about the results of each group.
  - It should work as a separate module of the main documentation generator.
 
- The sanity check should check
  - That there is no tag in the documentation that is missing in config.yml.
  - The required fields such as brief, metadata, and test are present as well as its subitems.

PR #1619, opened 2021-07-20
Add Config to DocGenerator

palaciosjeremias:
|Related issue|
|---|
|#1608|

## Description
This PR adds the module Config to the DocGenerator solution.
DocGenerator now will have a config.yaml file with all the configurable parameters and each of these configurations can be directly consulted to the module by public parameters.
The configurable values are:

- Project path. The complete QA path
- Documentation path. Where the output documentation is dumped
- Include paths. All the paths that will be parsed to collect the documentation.
- Include regex. Regex to match only the test files.
- Ignore paths. List of paths that requires to be ignored during the parsing.
- Valid tags. List of the valid tags to be used. For sanity check.
- Required fields. List of the minimum fields that the documentation must provide. For sanity check.
- Ignored fields. List of the fields that mustn´t be documented.

### DIFF
--- a/docs/DocGenerator/Config.py
+++ b/docs/DocGenerator/Config.py
@@ -1,0 +1,127 @@
+from collections import namedtuple
+import yaml
+
+CONFIG_PATH = "config.yaml"
+
+
+class Config():
+    def __init__(self):
+        self.project_path = "../.."
+        self.documentation_path = ".."
+        self.include_paths = []
+        self.include_regex = []
+        self.ignore_paths = []
+        self.valid_tags = []
+        self.required_fields = self.__fields()
+        self.ignored_fields = self.__fields()
+        with open(CONFIG_PATH) as fd:
+            self.__config_data = yaml.load(fd)
+
+        self.__read_project_path()
+        self.__read_documentation_path()
+        self.__read_include_paths()
+        self.__read_include_regex()
+        self.__read_ignore_paths()
+        self.__read_valid_tags()
+        self.__read_required_fields()
+        self.__read_ignored_fields()
+
+    def __read_project_path(self):
+        if 'Project path' in self.__config_data:
+            self.project_path = self.__config_data['Project path']
+
+    def __read_documentation_path(self):
+        if 'Documentation path' in self.__config_data:
+            self.documentation_path = self.__config_data['Documentation path']
+
+    def __read_include_paths(self):
+        if not 'Include paths' in self.__config_data:
+            raise Exception("Include paths are empty")
+        for include in self.__config_data['Include paths']:
+            if not 'path' in include:
+                raise Exception("One include path is missing")
+            element = self.__paths()
+            element.path = include['path']
+            if 'recursive' in include:
+                element.recursive = include['recursive']
+            self.include_paths.append(element)
+
+    def __read_include_regex(self):
+        if not 'Include regex' in self.__config_data:
+            raise Exception("Include regex is empty")
+        self.include_regex = self.__config_data['Include regex']
+
+    def __read_ignore_paths(self):
+        if 'Ignore paths' in self.__config_data:
+            self.ignore_paths = self.__config_data['Ignore paths']
+
+    def __read_valid_tags(self):
+        if 'Valid tags' in self.__config_data:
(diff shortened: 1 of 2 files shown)


docs/DocGenerator/Config.py
```diff
@@ -0,0 +1,127 @@
...
+for element in conf.required_fields.module:
+    print(element)
+for element in conf.required_fields.tests:
+    print(element)
+
+print(conf.ignored_fields.case_sensitive)
+for element in conf.ignored_fields.module:
+    print(element)
+for element in conf.ignored_fields.tests:
+    print(element)
```
MiguelazoDS, line 127: I would remove from the `conf` object creation to the end.

docs/DocGenerator/Config.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __read_include_paths(self):
+        if not 'Include paths' in self.__config_data:
+            raise Exception("Include paths are empty")
+        for include in self.__config_data['Include paths']:
+            if not 'path' in include:
+                raise Exception("One include path is missing")
+            element = self.__paths()
+            element.path = include['path']
+            if 'recursive' in include:
```
MiguelazoDS, line 45: How would be the behavior if there's not a `recursive` option?

docs/DocGenerator/config.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - Vulnerability Detector
+
+Required fields:
+  Module:
+    - Brief
+    - Metadata
+    - Metadata.Modules
+    - Metadata.Daemons
+    - Metadata.Operating System
+    - Metadata.Tiers
```
MiguelazoDS, line 32: Could we consider another format for this? **Brief**, **Metadata**, and **Tests** should be at the same level. 

Something like this: 
```
Required fields:
   - Brief
   - Metadata: [Modules, Daemons, Operating System, Tiers]
   - Tests: [Test Logic, Checks] 
```

docs/DocGenerator/Config.py
```diff
@@ -0,0 +1,127 @@
...
+                self.ignored_fields.case_sensitive = ignored_fields['Case Sensitive']
+
+    class __paths:
+        def __init__(self):
+            self.path = []
+            self.recursive = True
+
+    class __fields:
+        def __init__(self):
+            self.module = []
```
MiguelazoDS, line 93: Probably this would be affected if we change the format in the **config.yml**

docs/DocGenerator/config.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+Include paths:
+  - wazuh_db:
+    path: "../../tests/integration/test_wazuh_db"
+    recursive: false
+  - vulnerability_detector:
+    path: "../../tests/integration/test_vulnerability_detector"
+
+Include regex:
+  - "^test_.*py$|readme.md$"
```
MiguelazoDS, line 13: What about something like this:
```
Include regex: "^test_.*py$|readme.md$"
```

palaciosjeremias: It is contemplated to have more than one regex here.

docs/DocGenerator/Config.py
palaciosjeremias: Yes. Changing the format of the config.yaml requires to adapt the config module.
The idea of having a dedicated method for the configuration is to keep the module stable even if the config file is wrongly edited.

docs/DocGenerator/config.yaml
palaciosjeremias: Yes. This fields will be adapted in the Sanity module development ([link](https://github.com/wazuh/wazuh-qa/issues/1616))

docs/DocGenerator/Config.py
palaciosjeremias: Without `recursive` option, the parser will only look for the files at the the first level of this path.

docs/DocGenerator/Config.py
palaciosjeremias: This part of the code is implemented for development purposes and will be deleted when it is fully integrated, Being able to be tested in the full solution.

MiguelazoDS (approved):

Issue #1628, opened 2021-07-21
Implement a Sanity Check module for DocGenerator


palaciosjeremias:
As part of the epic [Overall improvements for integration tests documentation](https://github.com/wazuh/wazuh-qa/issues/1536) we need to implement the module that is in charge of checking the coverage and sanity of the documentation.

The requirements are:

- The coverage function must evaluate how much of the already implemented tests are explained in the documentation. 
  - It should display comprehensive details about the results of each group.
  - It should work as a separate module of the main documentation generator.
 
- The sanity check should check
  - That all the Mandatory fields exist on each documentation file.
  - A coverage calculus must be implemented.
  - A list with all the tags found should be displayed.

PR #1649, opened 2021-07-26
Implement a sanity check module for DocGenerator

palaciosjeremias:
|Related issue|
|---|
|#1628|

## Description
This PR adds the new module Sanity to **DocGenerator** tool.
At the moment, this module is executed independently from the main DocGenerator tool with `python Sanity.py`
This module checks if every Mandatory field is present in the documentation jsons, identifies the different tags found and counts every test and test module to obtain a coverage calculus.


docs/DocGenerator/Config.py
```diff
@@ -14,10 +14,14 @@ def __init__(self):
...
-        self.required_fields = self.__fields()
-        self.ignored_fields = self.__fields()
-        with open(CONFIG_PATH) as fd:
-            self.__config_data = yaml.load(fd)
+        self.module_fields = self.__fields()
+        self.test_fields = self.__fields()
+
+        try:
+            with open(CONFIG_PATH) as fd:
+                self.__config_data = yaml.load(fd)
```
MiguelazoDS, line 22: We changed this to safe_load() because load() is deprecated.

docs/DocGenerator/Sanity.py
```diff
@@ -0,0 +1,123 @@
...
+                    with open(os.path.join(root,test_file)) as fd:
+                        file_content = fd.read()
+                    module = ast.parse(file_content)
+                    functions = [node for node in module.body if isinstance(node, ast.FunctionDef)]
+                    for function in functions:
+                        for regex in function_regexes:
+                            if regex.match(function.name):
+                                self.project_tests = self.project_tests + 1
+
+
```
MiguelazoDS, line 75: Too many spaces.

docs/DocGenerator/Config.py
palaciosjeremias: Fixed!

docs/DocGenerator/Sanity.py
palaciosjeremias: Fixed

MiguelazoDS (approved):

Issue #1664, opened 2021-07-28
Implement the mandatory and optional fields feature on DocGenerator


palaciosjeremias:
As part of the Epic [1536](https://github.com/wazuh/wazuh-qa/issues/1536) we have implemented a tool capable to extract documentation blocks from the comments of the test.
We have to add the functionality of optional and mandatory fields.
- During the parse stage, every field not listed as mandatory or optional should be ignored.
- During the sanity stage, every mandatory field not found in the documentation should be reported.

PR #1665, opened 2021-07-28
Implement Mandatory and Optional fields functionality

palaciosjeremias:
|Related issue|
|---|
|#1664|

## Description
This PR adds a configurable way to decide which fields must be included in the documentation and which must be checked during a sanity scan.

Each field not found in the Mandatory or the Optional fields will be removed from the documentation output.
Each field in the Mandatory fields will be checked during the Sanity execution.
This is implemented for modules and tests separately.

A new Utils module was included with useful dictionary handling functions.
Also, some example documentation blocks were added to WazuhDB test.

### DIFF
--- a/docs/DocGenerator/CodeParser.py
+++ b/docs/DocGenerator/CodeParser.py
@@ -5,11 +5,13 @@ import json
 import yaml
 from Config import Config
 from TestCaseParser import TestCaseParser
+from Utils import remove_inexistent
 from docstring_parser import parse
 from comment_parser import comment_parser
 import warnings
 
-INTERNAL_FIELDS = ['Tests','Test Cases', 'Id', 'Group Id', 'Name']
+INTERNAL_FIELDS = ['Id', 'Group Id', 'Name']
+STOP_FIELDS = ['Tests','Test Cases']
 
 class CodeParser:
     def __init__(self):
@@ -27,14 +29,11 @@ class CodeParser:
 
     def remove_ignored_fields(self, doc):
         allowed_fields = self.conf.module_fields.mandatory + self.conf.module_fields.optional + INTERNAL_FIELDS
-        for field in list(doc):
-            if field not in allowed_fields:
-                del doc[field]
-        allowed_fields = self.conf.test_fields.mandatory + self.conf.test_fields.optional + INTERNAL_FIELDS
-        for test in list(doc['Tests']):
-            for field in list(test):
-                if field not in allowed_fields:
-                    del test[field]
+        remove_inexistent(doc, allowed_fields, STOP_FIELDS)
+        if 'Tests' in doc:
+            allowed_fields = self.conf.test_fields.mandatory + self.conf.test_fields.optional + INTERNAL_FIELDS
+            for test in doc['Tests']:
+                remove_inexistent(test, allowed_fields, STOP_FIELDS)
 
     def parse_comment(self, function):
         docstring = ast.get_docstring(function)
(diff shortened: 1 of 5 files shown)


docs/DocGenerator/config.yaml
```diff
@@ -6,8 +6,6 @@ Include paths:
   - wazuh_db:
     path: "../../tests/integration/test_wazuh_db"
     recursive: false
-  - vulnerability_detector:
```
MiguelazoDS, line 9: I think we should use the same separator. Underscore instead of space in all keys.

tests/integration/test_wazuh_db/test_wazuh_db.py
```diff
@@ -1,6 +1,29 @@
...
+    Operating System:
+        - Windows
+        - Ubuntu
+    Wazuh Max Version: 4.0.0
+    Wazuh Min Version: 4.1.5
+    Tiers:
+        - 0
+        - 1
+    Tags:
+        - Enrollment
```
MiguelazoDS, line 25: Since the tags are optional, I assume we are not going to validate them. Should be a problem to accept any word (or phrase?) as tags? I think about capital letters, misspelling words like  `enrrolment` instead of `Enrollment`, etc.

docs/DocGenerator/config.yaml
```diff
@@ -22,24 +20,15 @@ Ignore paths:
...
-      - Metadata:
-        - Modules
-        - Daemons
-        - Operating System
-        - Tiers
-        - Demo:
-          - One
-          - Two
+      - Metadata
+      - Modules
```
MiguelazoDS, line 28: Isn't it a little bit confusing to have the `metadata` fields out of it?

docs/DocGenerator/config.yaml
palaciosjeremias: Did you refer to the keys for other items in the config?
I think using spaces is more readable as it isn't required to be a unique word.
In this case, wazuh_db or vulnerability_detector refers to a Wazuh module.

tests/integration/test_wazuh_db/test_wazuh_db.py
palaciosjeremias: I agree. But, the requirement is to accept any tag found in the documentation. During the sanity check stage, all the tags found will be listed as a way to identify this kind of problem.
On the other hand, we can propose to move every tag to a specific case.

docs/DocGenerator/config.yaml
palaciosjeremias: This change was aimed to simplify the configuration file and make it more flexible. Having a tree of allowed values was more strict to write in the documentation and harder to parse in the code.
We will check if it is required to modify this.

palaciosjeremias: Now, the Optional and Mandatory fields have a tree format as suggested.

docs/DocGenerator/CodeParser.py
```diff
@@ -5,10 +5,14 @@
 import yaml
 from Config import Config
 from TestCaseParser import TestCaseParser
+from Utils import remove_inexistent
 from docstring_parser import parse
 from comment_parser import comment_parser
 import warnings
 
+INTERNAL_FIELDS = ['Id', 'Group Id', 'Name']
```
MiguelazoDS, line 13: Since we agreed to use snake case for those fields, we should change this.

docs/DocGenerator/CodeParser.py
```diff
@@ -5,10 +5,14 @@
...
+STOP_FIELDS = ['Tests','Test Cases']
```
MiguelazoDS, line 14: ... And this

docs/DocGenerator/CodeParser.py
palaciosjeremias: This changes were found and requested during the exploratory testing.
So they are included in this [PR](https://github.com/wazuh/wazuh-qa/pull/1680)

docs/DocGenerator/CodeParser.py
palaciosjeremias: Ditto

MiguelazoDS (approved):

Issue #1641, opened 2021-07-23
DocGenerator exploratory testing


palaciosjeremias:
As part of the Epic [1536 ](https://github.com/wazuh/wazuh-qa/issues/1536)we have implemented DocGenerator, a tool capable of finding test files and test functions to parse its documentation blocks comments. We need to perform a complete exploratory testing cycle to validate the correct behavior of the tool.
At least, the following items must be validated:
- Documentation content with text stylizers.
- Different options for the include paths and include regexes.
- Different options for the Ignore paths
- SanityCheck module identifies every Mandatory field missing
- CodeParser field bypass every field not mandatory or optional

palaciosjeremias:
During the exploratory testing the following errors were found that must be fixed:
- [ ] The default output folder isn´t under the git ignore list
- [ ] The configuration file refers to a deprecate option "recursive" for include and ignore paths
- [ ] The output folder isn´t cleaned on each parser run.
- [ ] Empty results after parsing are generating an empty file.
- [ ] Terminal output of the script is too verbose, which makes the warning messages to get lost

palaciosjeremias:
It was identified that the keys in the output structure must avoid white spaces to simplify the parsing into SearchUI. I will be better to:
- [] Change documentation key names to snake_case.

MiguelazoDS:
Some consideration that appears during the implementation of 
- [ ] Evaluate the possibility to move the helper modules to a `lib` directory. 
- [ ] Revise the function is_valid_file. Some conditions may never occur and the return value must be unassigned.

PR #1680, opened 2021-08-03
Exploratory testing fixes

palaciosjeremias:
|Related issue|
|---|
|#1641|

## Description
This PR include the fixes for the minor problems found during the exploratory testing:
 - The default output folder isn´t under the git ignore list
 - The configuration file refers to a deprecate option "recursive" for include and ignore paths
 - The output folder isn´t cleaned on each parser run.
 - Empty results after parsing are generating an empty file.
 - Terminal output of the script is too verbose, which makes the warning messages to get lost
 - A log module is required
 - Include path list is not clear enough as it requires a name to identify each element
 - Test Case functionality should be optional.
 - The tool must be documented with the same yaml format of the tests
 - The tool should support arguments to be called. i.e.: -s for a Sanity check or -d to enable the debug log level
 - Change documentation key names to snake_case.

### DIFF
--- a/docs/DocGenerator/CodeParser.py
+++ b/docs/DocGenerator/CodeParser.py
@@ -44,12 +44,9 @@ class CodeParser:
 
         except Exception as inst:
             if hasattr(function, 'name'):
-                warnings.warn(f"Error parsing comment of function '{function.name}'' from module {self.scan_file}")
+                warnings.warn(f"Error parsing comment of function '{function.name}'' from module {self.scan_file}", stacklevel=2)
             else:
-                warnings.warn(f"Error parsing comment of module {self.scan_file}")
-            print(type(inst))
-            print(inst.args)
-            print(inst)
+                warnings.warn(f"Error parsing comment of module {self.scan_file}", stacklevel=2)
             doc = None
 
         return doc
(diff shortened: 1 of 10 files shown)


docs/DocGenerator/CodeParser.py
```diff
@@ -1,92 +1,147 @@
...
-        self.test_case_parser = TestCaseParser()
+        self.pytest = PytestWrap()
         self.function_regexes = []
         for regex in self.conf.function_regex:
             self.function_regexes.append(re.compile(regex))
 
     def is_documentable_function(self, function):
+        """
+        brief: Checks if a specific method match with the regexes to be documented.
+        args: -"function (_ast.FunctionDef): Function class with all the information of the method"
```
MiguelazoDS, line 41: Is there a reason we put the elements of `args` in the same line?

MiguelazoDS (approved):

docs/DocGenerator/CodeParser.py
palaciosjeremias: No. It´s only because it was a single element. But we can use an extra line if you consider it will be better presented.

PR #1700, opened 2021-08-10
Fixes for QACTL

### LINKED ISSUE #1695: QACTL fixes
After testing the development of the QACTL we found the following issues:

Right now, we have a first working tool in the following branch: https://github.com/wazuh/wazuh-qa/tree/qa-ctl
We have found the following issues that need to be fixed in this first iteration:

1. [x] Docker assigns IP dynamically.
2. [x] When a VM is destroyed, the SSH fingerprint is not removed from `known_hosts`, causing the following failure:
 ```
TASK [Create None/wazuh path] **************************************************
fatal: [10.2.0.3]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: 
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the ED25519 key sent by the remote host is
SHA256:O7NzwSNf0OvmtyQajn74u7ktz7177CvkJ90LfuAq68k.
Please contact your system administrator.
Add correct host key in /home/antoniomanuelfr/.ssh/known_hosts to get rid of this message.
Offending ECDSA key in /home/antoniomanuelfr/.ssh/known_hosts:10
Password authentication is disabled to avoid man-in-the-middle attacks.
Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
UpdateHostkeys is disabled because the host key is not trusted.
vagrant@10.2.0.3: Permission denied (publickey,password).", "unreachable": true}
 ```

3. [x] Finish testing the test module.
4. [x] If a box deployed with qa-ctl has the same name that an existing one, an exception is raised aborting the execution.
This one is a bit problematic, as the `python-vagrant` module is doing all the operations calling using the `subprocess` command, and when the command fails, the only information that is returned is that it returned with a non-zero exit status, but not the reason why the command failed. 
5. [x] Also, the containers deployed with docker are also failing when an existing container has the same name as the one deployed with qa-ctl
6. [x] To download installation sources files, we are doing a git clone. This is not efficient enough and also generates problems when the repository is already cloned before. Instead, we will download the files directly.
7. [x] Remove vagrantfile folder when destroying the box
8. [x] Some minor problems regarding ossec.log and ossec.conf not showing as plain text in browser. It shows as URL data with the following scheme: data:<mediatype>:<base64 encoded data>. Maybe it needs to be changed to fetch the assets folder as well instead of using --self-contained-html flag.
9. [x] The remove option in the docker section is not taking the default value

Implement WazuhS3Package class to allow package downloads from Wazuh packages repository, this will be added in this issue: https://github.com/wazuh/wazuh-qa/issues/1701
It would be a good idea to show all the pytest output when running test, to see the progress, and if any of them have failed or not. (Edit: This can't be completed, explication below in the comments)

The purpose of this issue is to collect all the detected problems and fix them, so the above list may increase while we detect new problems

antoniomanuelfr:
|Related issue|
|---|
|#1695|

## Description
This issue aims to add the fixes described in #1695.

Closes #1695

### DIFF
--- a/deps/conf.yaml
+++ b/deps/conf.yaml
@@ -13,6 +13,20 @@ deployment:
                 label: 'Label'
                 vm_ip: '10.2.0.13'
                 quiet_out: False
+    host2:
+        provider:
+            vagrant:
+                enabled: True
+                vagrantfile_path: '/tmp'
+                vagrant_box: '/home/vhalgarv/drawers/qactlubuntu20.box'
+                vm_memory: 2048
+                vm_cpu: 2
+                vm_name: 'testqactlubuntu2'
+                vm_system: 'Linux'
+                os: Linux
+                label: 'Label'
+                vm_ip: '10.2.0.12'
+                quiet_out: False
 provision:
     hosts:
         host1:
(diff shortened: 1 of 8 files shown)


jotacarma90 (approved):
LGTM

PR #1739, opened 2021-08-19
Change HTML report path

### LINKED ISSUE #1728: QACTL: Fix HTML report generation and update documentation
## Description
When test are run, the HTML report file creation makes that program exits with `File not found error`

Also, the QACTL Test Execution [documentation](https://github.com/wazuh/wazuh-qa/issues/1533#issuecomment-895019665) is outdated, we need to check and update it.
## Tasks
- [x] HTML report is not generated correctly.
- [x] HTML report does not contain test info
- [x] Check and update documentation.
## Additional task
- [x] Use a new directory inside tests execution dir, in order to not fill the directory with assets, plain and html reports

roronoasins:
|Related issue|
|---|
|close #1728|

## Description

This PR makes the followings changes:

- Fix HTML report file and assets path

## Checks
HTML report correctly working now as showed here https://github.com/wazuh/wazuh-qa/issues/1728#issuecomment-901797765

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/qa_ctl/run_tests/Pytest.py
+++ b/deps/wazuh_testing/wazuh_testing/qa_ctl/run_tests/Pytest.py
@@ -71,19 +71,23 @@ class Pytest(Test):
         Args:
             ansible_inventory_path (str): Path to ansible inventory file
         """
-
+        # Paths used below
         assets_folder = 'assets/'
         assets_zip = "assets.zip"
-        html_report = "/home/vagrant/"
+        html_report_file_name = f"test_report-{datetime.now()}.html"
+        plain_report_file_name = f"plain_report-{datetime.now()}.txt"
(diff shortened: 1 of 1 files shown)


deps/wazuh_testing/wazuh_testing/qa_ctl/run_tests/Pytest.py
```diff
@@ -74,6 +74,7 @@ def run(self, ansible_inventory_path):
 
         assets_folder = 'assets/'
         assets_zip = "assets.zip"
+        html_report = "/home/vagrant/"
```
jmv74211, line 77: Be careful with paths ended in `/`. Its really necessary?

jmv74211: You cannot hardcode this path, note that this will not only run on machines deployed with vagrant.

jmv74211 (changes requested):
Good job, but you need to make some changes

jmv74211 (approved):

Issue #1724, opened 2021-08-17
Remote upgrading for MacOS agents


TomasTurina:
### Description

Support for remote upgrading of MacOS agents needs to be considered in WPK integration tests.

Some things to consider:

- Remove MacOS from case of failure in WPK manager tests.
- Adapt WPK agent tests to be able to run on MacOS agents.

snaow:
Closed by https://github.com/wazuh/wazuh-qa/pull/1853

PR #1745, opened 2021-08-20
Enable WPK tests for macOS agents

TomasTurina:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-qa/issues/1724|

## Description

Remote upgrading using WPK files now supports MacOS agents.

Following this new feature, this changes have been included in the WPK integration tests:

- Manager:
  - Replace unsupported OS from `mojave` to `solaris11`.
  - Run custom upgrade tests with `mojave` agents.
- Agent (MacOS):
  - Run agent tests on MacOS agents.

## Tests

- [x] The test is documented in wazuh-qa/docs.


Rebits:
Due to new repository standards, the simulators tools (Authd, Remoted and Agent Simulators) will be deprecated. For this reason, every PRs that changed any test that uses then should be reconsidered as follows:

- Perform this check without using any simulator. 
- If it is impossible for an integration test to check this case, a system test should be created. This will be noted as future work in a common issue for each module.

TomasTurina:
Moved to https://github.com/wazuh/wazuh-qa/pull/1853

PR #1753, opened 2021-08-23
Add files and local internal options to the HTML report after executing a test

### LINKED ISSUE #1690: Modify the QA framework to the save the logs after every test
### Description

When a test fails, sometimes it is difficult to find out the reason because the QA framework doesn't save the `ossec.log` and `archives.log` files. These logs are truncated for every test and that information isn't available when it fails.

It would be really helpful during the debugging process to have them, so it isn't necessary to recover them manually.

MizugorouZ:
|Related issue|
|---|
|Closes #1690 |

## Description
This PR makes possible to add several files to the HTML report generated when executing a test. When the test failed, the specified files are linked in a similar way the configuration files `ossec.log` and `ossec.conf` are linked.

The files mentioned above are always linked when the HTML report is generated. even when not specified by the user. In addition to those files, we think is a good idea to add `local_internal_options.conf` aswell.

The ussage of this feature would be as follows:
`python3 -m pytest test --html=/path/to/report.html --add-file=/path/to/file1 --add-file=/path/to/file2 --add-file=...`

## Tests results
[This](https://github.com/wazuh/wazuh-qa/files/7031865/r4_ejemplo.zip) is an example of the execution of a failed test which will add 3 files to the report, aswell as the configuration files.


tests/integration/conftest.py
```diff
@@ -311,7 +334,8 @@ def pytest_runtest_makereport(item, call):
         extra.append(pytest_html.extras.json(arguments, name="Test arguments"))
 
         # Extra files to be added in 'Links' section
-        for filepath in (LOG_FILE_PATH, WAZUH_CONF):
+        files = get_report_files()
+        for filepath in files:
```
damarisg, line 338: When I add a file that not exists the all execution fail. 
I executed: 
 `python3 -m pytest -svx test_vulnerability_detector/test_feeds/ --html=/vagrant/r1_test.html --add-file=/var/ossec/logs/archives/archives.json > /vagrant/LogsConFails.log`
I Attach results [Logs.zip](https://github.com/wazuh/wazuh-qa/files/7086637/SampleOfFileNotExists.zip). 

For this case, it also should work without problems. (only the file will do not attached to the log)

tests/integration/conftest.py
```diff
@@ -180,6 +192,14 @@ def pytest_addoption(parser):
         type=str,
         help="run tests using a specific WPK package version"
     )
+    parser.addoption(
+        "--add-file",
+        action="append",
+        metavar="add-file",
+        default=[],
+        type=str,
+        help="add file to the HTML report"
```
damarisg, line 201:
```suggestion
        help="store logs to the HTML report"
```

tests/integration/conftest.py
```diff
@@ -180,6 +192,14 @@ def pytest_addoption(parser):
         type=str,
         help="run tests using a specific WPK package version"
     )
+    parser.addoption(
+        "--add-file",
+        action="append",
+        metavar="add-file",
```
damarisg, line 198:
```suggestion
        metavar="log",
```

damarisg: I think that name should be clear and short. 
I suggest `log` but it only an idea.

damarisg (changes requested):
I would like that the results always to return the files  `ossec.conf`, `ossec.log`, and `local_internal_option.conf` not only when the test fails.

tests/integration/conftest.py
Deblintrake09: maybe log-file would be a more suitable name, since it adds a file that would serve as part of the log, but is not necesarily a .log per se (we could pull a conf, json, txt, etc).

```suggestion
        metavar="log-file",
```

Deblintrake09:
Minor changes that cannot be requested as changes for the PR:
 
```Line 122: 
                        help="only run tests with a tier level less or equal than 'minimum_level'"
                        help="only run tests with a tier level greater or equal than 'minimum_level'"
```

```Line 130: 
                        help="only run tests with a tier level less or equal than 'minimum_level'"
                        help="only run tests with a tier level less or equal than 'maximum_level'"
```

Deblintrake09:
**Tests Results:**

Test 1:  `python3 -m pytest -svx test_vulnerability_detector/test_feeds/ --html=/vagrant/PR1753.html --add-file=/var/ossec/logs/archives/archives.log > /vagrant/PR1753.log`
[PR1753.zip](https://github.com/wazuh/wazuh-qa/files/7091783/PR1753.zip)



Test 2: `python3 -m pytest -svx test_logtest/test_rules_decoders_load/ --html=/vagrant/PR1753-2.html --add-file=/var/ossec/ruleset/decoders/0090-dragon-nids_decoders.xml --add-file=/var/ossec/ruleset/decoders/0070-cisco-vpn_decoders.xml > /vagrant/PR1753-2.log`
[PR1753-2.zip](https://github.com/wazuh/wazuh-qa/files/7091785/PR1753-2.zip)

Deblintrake09 (approved):
**Tests Results:**

Test 1:  `python3 -m pytest -svx test_vulnerability_detector/test_feeds/ --html=/vagrant/PR1753.html --add-file=/var/ossec/logs/archives/archives.log > /vagrant/PR1753.log`

[PR1753.zip](https://github.com/wazuh/wazuh-qa/files/7091814/PR1753.zip)




Test 2: `python3 -m pytest -svx test_logtest/test_rules_decoders_load/ --html=/vagrant/PR1753-2.html --add-file=/var/ossec/ruleset/decoders/0090-dragon-nids_decoders.xml --add-file=/var/ossec/ruleset/decoders/0070-cisco-vpn_decoders.xml > /vagrant/PR1753-2.log`

[PR1753-2.zip](https://github.com/wazuh/wazuh-qa/files/7091817/PR1753-2.zip)

damarisg (approved):

MizugorouZ:
The following examples are the results of 2 tests forced to fail. Both HTML have been generated with `--self-contained-html`. The first one contains an example file attached, while the other was requested to have a file that doesn't exist, showing in the report there is no file added:
| Results | Date | Executed by | Status |
|:--:|:--:|:--:|:--:|
| [Archivo existente](https://github.com/wazuh/wazuh-qa/files/7143335/ejemplo_self_contained.zip) | 2021/09/10 | @MizugorouZ | :green_circle: |
| [Archivo NO existente](https://github.com/wazuh/wazuh-qa/files/7143336/ejemplo_self_contained_no_existe.zip) | 2021/09/10 | @MizugorouZ | :green_circle: |

PR #1897, opened 2021-09-15
Remove the disk i/o % usage calculation from the performance tools

### LINKED ISSUE #1718: Incorrect disk usage calculation in performance tools
## Description

It seems that the disk usage calculation is wrong in the `wazuh_testing/tools/performance/binary.py` script. If we observe any disk graphics created by said tool, the following behavior appears: 

imagen

Apart from the fact that disk growth follows a logarithmic trend that is strange, we can see that it far exceeds 100% of use, which should not be possible. This is done to calculate the percentage of disk I/O:
https://github.com/wazuh/wazuh-qa/blob/ef1723d37835705eeceeb1a7f2abc0c1ef4ceec1/deps/wazuh_testing/wazuh_testing/tools/performance/binary.py#L137-L146
1. Add the total number of `read_bytes` and `write_bytes` (both cumulative, according to the [psutil doc](https://psutil.readthedocs.io/en/latest/#psutil.Process.io_counters)) of the current process (#L139). 
2. Add the total number of `read_bytes` and `write_bytes` of the entire disk (here the [psutil doc](https://psutil.readthedocs.io/en/latest/#psutil.disk_io_counters) makes no mention of it being cumulative) 
3. Divide the result obtained in `1.` by that obtained in `2.` 

The number of read + writes of the process should never exceed the read + writes of the disk as a whole. That is the reason why it is strange that disk usage exceeds 100%. 

On the other hand, if we see, for example, the number of reading operations of the following graph (a test that lasted one day and in which the content of all agent-groups was updated every 30 seconds), we see that the number of KBs read remains static all the time, it never reads anything new, which also does not make sense and seems to indicate that something it is not working well: 
imagen

It would be necessary to review what the expected behavior is and if it coincides with the current one.

mcarmona99:
|Related issue|
|---|
| #1718 |

## Description

This PR closes #1718.

After the investigation done in the related issue, the decision taken was to delete the `disk(%)` measure from the process metrics.

This pull request removes this metric.

I have tested this in our testing environment and the plots generated are the ones including: disk read and written KBs, disk read and write speed; and read and write ops. (Talking about plots showing disk information)

The plot including `disk(%)` will no longer be created.

## Checks

- [X] `provision_documentation.sh` generate the docs without errors.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/performance/binary.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/performance/binary.py
@@ -102,7 +102,6 @@ class Monitor:
             - Write_Ops: write operations.
             - Disk_Read: Bytes read by the process.
             - Disk_Written: Bytes written by the process.
-            - Disk(%): percentage of the I/O operations ran by the process compared with the OS.
 
         Args:
             proc (psutil.proc): psutil object with the data of the process.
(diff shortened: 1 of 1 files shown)


AdriiiPRodri (approved):

davidjiglesias (approved):

Rebits (approved):
LGTM

Rebits:
Disk i/o % usage has been removed correctly -  [wazuh-analysisd](https://github.com/wazuh/wazuh-qa/files/7216688/wazuh-analysisd.csv)

jmv74211 (approved):

Issue #1877, opened 2021-09-13
qa-docs: Automatize SearchUI dependencies installation


roronoasins:
## Description
Task from #1864.

Currently, when you want to run SearchUI using `qa-docs -I tests_dir_path -l index_name`, you need to do a previous dependencies installation via `npm install`.

We need to automatize that installation, so the user doesn't need to worry about dependencies.

We have two installation options:
- Via framework `setup.py`
- When you run SearchUI, check that `node_modules/` exists. If not, install the dependencies

## Branch
[1877-deps-installation](https://github.com/wazuh/wazuh-qa/tree/1877-deps-installation)

## Tasks
- [x] Check that `node_modules/` exists before launching `SearchUI`
- [x] Install the dependencies if it is necessary
- [x] merge `1877-deps-installation` to `1864-qa-docs-fixes`

PR #1968, opened 2021-10-04
Install the SearchUI dependencies if necessary.

roronoasins:
|Related issue|
|---|
|#1877|

## Description
With these changes, when running SearchUI its dependencies are installed(if necessary) so you do not need to install them manually.

# Checks


mdengra (approved):
Great work!

Issue #1943, opened 2021-09-29
QA-CTL Improvements


fernandolojano:
This issue has been opened to improve some features for the `qa-ctl` tool.
As for now, all the files are created in the temporary directory of the system where the tool is running, but we need these files to be created in a new folder called `qa-ctl` so we can keep all of them organized in this new folder and clean up a little bit the temporary system folder. 
Another needed improvement is to check that the `qa-ctl` process is stopped correctly if some of the tool steps fail. It seems like the current stop implementation is not working properly.

fernandolojano:
The working branch for this issue is : [1943-qa-ctl-improvements](https://github.com/wazuh/wazuh-qa/tree/1943-qa-ctl-improvements)

fernandolojano:
As for now, the first part of the issue, which is storing all the temporary files created by `qa-ctl` in a single folder to keep the temporary files system folder organised is done and solved.
After making some research about the second issue, I found that python threads not stopping after pressing `ctrl+c` (for stopping a process) is pretty common. However, it does not have a clear solution, there are multiple approaches instead. I'm currently trying to kill a python thread with the second approach suggested in this [page](https://blog.miguelgrinberg.com/post/how-to-kill-a-python-thread), that is, by using `Python Events`.

PR #1993, opened 2021-10-06
QA-CTL: Add new folder level for qa-ctl temporary files

fernandolojano:
|Related issue|
|---|
|issue #1943  |

## Description
The current `qa-ctl` implementation generates all its temporary files in the correspondent system temp folder. However, it is needed a better organization of these files so they can be tracked in an easier way. In order to achieve this purpose, we are going to modify the current implementation and add a new folder level called `qa_ctl`. 
This PR makes the following changes:
- Add a new folder level for temporary files called `qa_ctl`
- Modifies the file `vagrantfile_template.txt` to avoid updating the `VirtualBox guest additions`

## Checks
All the `qa-ctl` generated files are now stored in a folder called qa_ctl located in the temporary system folder. Plus, If the `vbguest` plugin is present in the system, the `VirtualBox guest additions` won't be updated during the `qa-ctl` execution.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/scripts/qa_ctl.py
+++ b/deps/wazuh_testing/wazuh_testing/scripts/qa_ctl.py
@@ -26,7 +26,7 @@ from wazuh_testing.tools.file import recursive_directory_creation
 DEPLOY_KEY = 'deployment'
 PROVISION_KEY = 'provision'
 TEST_KEY = 'tests'
-WAZUH_QA_FILES = os.path.join(gettempdir(), 'wazuh-qa')
+WAZUH_QA_FILES = os.path.join(gettempdir(), 'qa_ctl', 'wazuh-qa')
 
 qactl_logger = Logging(QACTL_LOGGER)
 _data_path = os.path.join(os.path.dirname(os.path.dirname(os.path.realpath(__file__))), 'data')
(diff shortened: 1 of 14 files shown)


deps/wazuh_testing/wazuh_testing/scripts/qa_ctl.py
```diff
@@ -113,7 +114,8 @@ def set_environment(parameters):
     """
     if parameters.run_test:
         # Download wazuh-qa repository locally to run qa-docs tool and get the tests info
-        local_actions.download_local_wazuh_qa_repository(branch=parameters.qa_branch, path=gettempdir())
+        recursive_directory_creation(os.path.join(gettempdir(), 'qa_ctl'))
```
jmv74211, line 117: Maybe this should be outside the `if` (find a good place to put it), because in the case of manual execution, this directory would not be created and would fail.

jmv74211 (changes requested):

deps/wazuh_testing/wazuh_testing/scripts/qa_ctl.py
fernandolojano: This change has been made in the commit [8e0c857](https://github.com/wazuh/wazuh-qa/commit/8e0c8573d76c0285c5bd9a388bc07ebdfb3bccfe). The folder creation process is located now in the main function of `qa_ctl.py`. With this solution, the qa_ctl folder will be always created, either in the case of an automated execution or a manual one.

jmv74211 (approved):

Issue #2053, opened 2021-10-15
New Authd force options test design


palaciosjeremias:
As part of https://github.com/wazuh/wazuh-qa/issues/2050 we need to design a new test that checks all the possible configurations of the force block that satisfies the new QA standards.
We need to:
- Define the documentation schema to be used.
- Define how Authd will restart along the tests to implement the different configuration
- Define how to test scenarios that require pre-existent agents.
- Define how to override the configurations along with the different tests.
- Implement a PoC that will be validated by the QA team.

PR #2055, opened 2021-10-15
force_options tests design

palaciosjeremias:
|Related issue|
|---|
|#2053|

## Description
This PR implements a PoC as the result of the design task for the new Authd force_options tests.
- Uses the DocGenerator 2.0 schema (https://github.com/wazuh/wazuh-qa/issues/1694).
- Makes a restart of Authd for each test case to set up the configuration of the test case.
- Inserts agents pre-installed by writing in **client.keys** and **global.db** on each test case, previous restarting Authd.
- For each test case, the configuration is overridden with the configuration of the test case.


tests/integration/test_authd/force_options/data/test_cases/after_registration_time.yaml
```diff
@@ -0,0 +1,70 @@
+-
```
Rebits, line 1: It should be useful to add an id field without spaces and symbols. This will help us to Individually identify every tcase

tests/integration/test_authd/force_options/data/test_cases/disconnected_time.yaml
```diff
@@ -0,0 +1,70 @@
+-
```
Rebits, line 1: Same

tests/integration/test_authd/force_options/data/test_cases/legacy_force_insert.yaml
```diff
@@ -0,0 +1,33 @@
+-
```
Rebits, line 1: Same

tests/integration/test_authd/force_options/data/test_cases/key_mismatch.yaml
```diff
@@ -0,0 +1,70 @@
+-
```
Rebits, line 1: Same

tests/integration/test_authd/force_options/test_authd_force_options.py
```diff
@@ -0,0 +1,201 @@
...
+monitored_sockets_params = [('wazuh-authd', None, True), ('wazuh-db', None, True)]
+receiver_sockets, monitored_sockets, log_monitors = None, None, None  # Set in the fixtures
+
+# Functions
+def validate_authd_logs(logs):
+    for log in logs:
+        log_monitor.start(timeout=AUTHD_KEY_REQUEST_TIMEOUT,
+                          callback=make_callback(log, prefix=AUTHD_DETECTOR_PREFIX,
+                                                 escape=True),
+                          error_message='Expected error log does not occured.')
```
Rebits, line 44: This could be useful not only for authd tests. Maybe it could be a good idea making it more general and move it to monitoring module.

tests/integration/test_authd/force_options/test_authd_force_options.py
```diff
@@ -0,0 +1,201 @@
...
+        if legacy_force_insert_conf:
+            temp_conf_file[0]['sections'][0]['elements'].append(legacy_force_insert_conf)
+    with open(temp, 'w') as temp_file:
+        yaml.safe_dump(temp_conf_file, temp_file)
+    return temp
+
+
+# Fixtures
+
+@pytest.fixture(scope='module', params=configurations, ids=configuration_ids)
```
Rebits, line 76:
```suggestion
@pytest.fixture(scope='module')
```

tests/integration/test_authd/force_options/test_authd_force_options.py
```diff
@@ -0,0 +1,201 @@
...
+            force_conf['force']['elements'].append({elem: {'value': param[elem]}})
+
+    with open(configurations_path, 'r') as conf_file:
+        temp_conf_file = yaml.safe_load(conf_file)
+        temp_conf_file[0]['sections'][0]['elements'].append(force_conf)
+        if legacy_force_insert_conf:
+            temp_conf_file[0]['sections'][0]['elements'].append(legacy_force_insert_conf)
+    with open(temp, 'w') as temp_file:
+        yaml.safe_dump(temp_conf_file, temp_file)
+    return temp
```
Rebits, line 71: IMHO this function should be split into the following function/fixtures:
- create_force_config_block: Create the config block following configure_environment fixture logic:
```
[{
   "section":"auth",
   "elements":[
      {
         "disabled":{
            "value":"no"
         }
      },
      {
         "port":{
            "value":1515
         }
      },
      {
         "use_source_ip":{
            "value":"no"
         }
      },
      {
         "force_insert":{
            "value":"yes"
         }
      },
      {
         "force_time":{
            "value":0
         }
      },
      {
         "purge":{
            "value":"yes"
         }
      },
      {
         "use_password":{
            "value":"no"
         }
      },
      {
         "limit_maxagents":{
            "value":"yes"
         }
      },
      {
         "ciphers":{
            "value":"HIGH:!ADH:!EXP:!MD5:!RC4:!3DES:!CAMELLIA:@STRENGTH"
         }
      },
      {
         "ssl_verify_host":{
            "value":"no"
         }
      },
      {
         "ssl_manager_cert":{
            "value":"/var/ossec/etc/sslmanager.cert"
         }
      },
      {
         "ssl_manager_key":{
            "value":"/var/ossec/etc/sslmanager.key"
         }
      },
      {
         "ssl_auto_negotiate":{
            "value":"no"
         }
      }
   ]
}]
```
- override_authd_force_conf (existing fixture): Use `configuration.set_section_wazuh_conf` function to set  configuration. Also, this fixture should backup the existing configuration using  `conf.get_wazuh_conf()` and set it after yield as follows:
```
    yield
    configuration.write_wazuh_conf(backup_config)
```

tests/integration/test_authd/force_options/test_authd_force_options.py
```diff
@@ -0,0 +1,201 @@
...
+    test_name = request.node.originalname
+    configuration = get_current_test_case.get('configuration', {})
+    # Configuration for testing
+    temp = get_temp_force_config(configuration)
+    conf = load_wazuh_configurations(temp, test_name, )
+    os.remove(temp)
+
+    test_config = set_section_wazuh_conf(conf[0]['sections'])
+    # Set new configuration
+    write_wazuh_conf(test_config)
```
Rebits, line 106: Applying the last comment, this fixture can be generalized and moved to the integration tests conftest file.
In order to do that, create a fixture, `format_configuration` as follows:
```
@pytest.fixture(scope='function')
def format_configuration(get_current_test_case)
```
This, should format and return authd configuration according to the current standard.
Using this structure, generic override_authd_force_conf should be something similar to this:
```
 def override_authd_force_conf(format_configuration, request)
```

Another option is to change _`yaml`_ tcase in order to follow the current standard.

Rebits (approved):
I think the approach to these new tests is really good. Only some minor aspects should be standardized in order to create a common structure for future tests

Issue #2167, opened 2021-11-03
Adapt WazuhDB tests to handle the new disconnection_time column in global DB


palaciosjeremias:
As part of https://github.com/wazuh/wazuh-qa/issues/2050, we need to add disconnection_time as part of the response that WazuhDB returns.
The core development added this new column and some tests ask for the complete global DB content, which makes now these tests fail.

DProvinciani:
Closing the issue as the pull https://github.com/wazuh/wazuh-qa/pull/2170 was merged.

PR #2170, opened 2021-11-03
Fix wdb tests with disconnection_time

palaciosjeremias:
|Related issue|
|---|
|#2167|

## Description
This PR adds the disconnection_time check on the respective **Wazhuh-DB** tests.
When the value is expected to be in 0 (like in the **never_connected** status), it is checked that this value is 0.
When the value can be any value based on the system time (like in the **disconnected** status), it is checked that this value is returned, but the value is bypassed with the wildcard *****

## Tests

- [X] `provision_documentation.sh` generate the docs without errors.


palaciosjeremias:
Succesful execution of WazuhDB after the changes.
image

DProvinciani (approved):
Looks good!

Issue #2373, opened 2021-12-21
Add test cases for IPv6 to Remoted tests - Local IP valid


MizugorouZ:
| Related issue |
|:--:|
| #2257 |

### Description
After the research done in #2257, `test_basic_configuration_local_ip_valid.py` needs to add cases to support IPv6.

Rebits:
Merged in https://github.com/wazuh/wazuh-qa/pull/2439

Rebits:
Due to https://github.com/wazuh/wazuh-qa/pull/2379#issuecomment-1005640450, `local_ip` test will not be ampliated for ipv6 tests cases for the moment.

PR #2379, opened 2021-12-22
IPv6 integration tests: Remoted local IP valid test

MizugorouZ:
|Related issue|
|---|
|#2373|

## Description
Following issue #2373, it is needed to add IPv6 cases to Remoted `test_basic_configuration_local_ip_valid.py`. We need to add the IPv6 parameter to the configuration file so we can test IPv4 and IPv6 by specifying the parameter to 'no' and 'yes'. In addition, we added the IPv6 address from all interfaces. There are test cases for every IPv4 and IPv6 address in each interface, with the ipv6 value to 'no' and 'yes' to test all possible scenarios.
```
- section: remote
    ...
    - ipv6:
        value: IPV6
    ...
```

## Tests
- [ ] Python codebase satisfies PEP-8 style style guide. `pycodestyle --max-line-length=120 --show-source --show-pep8 file.py`.
- [ ] Python codebase is documented following the [QA-Docs Schema 2.0](https://github.com/wazuh/wazuh-qa/wiki/QA-Documentation---How-to-document-a-test-using-Schema-2.0).

### DIFF
--- a/tests/integration/test_remoted/test_configuration/data/wazuh_basic_configuration.yaml
+++ b/tests/integration/test_remoted/test_configuration/data/wazuh_basic_configuration.yaml
@@ -61,6 +61,8 @@
         value: secure
     - port:
         value: 1514
+    - ipv6:
+        value: IPV6
     - local_ip:
         value: LOCAL_IP
     - protocol:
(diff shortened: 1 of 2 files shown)


Rebits (approved):
LGTM

Rebits:
After talking with @cborla and @Nicogp, it seems that for now, `local_ip` is only supported for global ips. Maybe link-local will be supported in the future, but for now, this implementation could cause false positives in the Jenkins environment. So it is required to change the way of cases (all valid local ips) is gathered.

Issue #2444, opened 2022-01-14
Implement an option to store the logs instead of deleting them in the system tests


yanazaeva:
Currently, at the beginning of every system test execution, the log files (`ossec.log` and `cluster.log`) are being cleared out. This can be inconvenient if we run the same test several times, and one of the execution fails, as we won't be able to check the logs. 

This kind of behavior can be seen in the[ following scenario](https://github.com/wazuh/wazuh-qa/pull/2411#issuecomment-1012398785), where a test failed due to some misfunctioning, but we cannot be sure what exactly it was.

PR #2445, opened 2022-01-14
Add logs to html report for system tests

juliamagan:
|Related issue|
|---|
|Closes: #2444 |

## Description

When a system test fails, there is no way to check the logs after the test is finished because they are deleted. In the future, when we adapt the system test to use [QA-CTL tool](https://github.com/wazuh/wazuh-qa/wiki/QACTL-tool), we'll need to change this method, but until then, to simplify the process of debugging an error, we have:
-  Modified `clean_tmp_files` from `HostMonitor` to copy the logs in `/tmp`.
- Create a `conftest.py` for system tests that adds those logs to the html report, and then deletes them from `/tmp`

When a test fails:

- If any logs are monitored, they are added to the report.
- If the test is part of a cluster, regardless of the monitored files, the cluster logs will also be added.

I attach two examples of the generated report:

| 🟢 | 🔴 |
|:--:|:--:|
|  [passed_report.zip](https://github.com/wazuh/wazuh-qa/files/10558365/2444-passed-report.zip)|  [failed_report.zip](https://github.com/wazuh/wazuh-qa/files/10558366/test_report.zip) |


tests/system/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+        extra.append(pytest_html.extras.json(arguments, name="Test arguments"))
+
+        # Extra files to be added in 'Links' section
+        files = get_report_files()
+        for filepath in files:
+            if os.path.isfile(filepath):
+                with open(filepath, mode='r', errors='replace') as f:
+                    content = f.read()
+                    extra.append(pytest_html.extras.text(content, name=os.path.split(filepath)[-1]))
+                os.remove(filepath)
```
Rebits, line 117: This should be a separate fixture.

tests/system/conftest.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (C) 2015-2021, Wazuh Inc.
```
Rebits, line 1: Almost all of the content of this file is the same as the integration test conftest. I suggest moving repeated code to a general conftest for all the tests.

deps/wazuh_testing/wazuh_testing/tools/monitoring.py
```diff
@@ -993,6 +993,11 @@ def clean_tmp_files(self):
         """Remove tmp files."""
         logger.debug(f'Cleaning temporal files...')
         for file in os.listdir(self._tmp_path):
+            if '.log.tmp' in file:
+                with open (os.path.join(self._tmp_path, file), 'r') as log:
+                    content=log.read()
+                with open (f"/tmp/{file.split('.')[0]}.txt", 'a') as failed_log:
+                    failed_log.write(content)
```
Rebits, line 1000: Why not use `copyfile` from `shutil`?

deps/wazuh_testing/wazuh_testing/tools/monitoring.py
```diff
@@ -993,6 +993,11 @@ def clean_tmp_files(self):
         """Remove tmp files."""
         logger.debug(f'Cleaning temporal files...')
         for file in os.listdir(self._tmp_path):
+            if '.log.tmp' in file:
```
Rebits, line 996:
```suggestion
            if '.log.tmp' in file:
            if file.endswith('.log.tmp'):
```

Rebits (changes requested):
GJ. Some changes are required

PR #2631, opened 2022-03-03
Update CSV Parser tool

### LINKED ISSUE #2468: Update cluster CSV parser tool and test thresholds
In https://github.com/wazuh/wazuh-qa/pull/2032, a tool was developed in charge of iterating, loading and processing CSVs with data on the use of resources of the `wazuh-clusterd` process.

However, multiple developments (https://github.com/wazuh/wazuh/pull/10767, https://github.com/wazuh/wazuh/pull/10807, https://github.com/wazuh/wazuh/issues/10920, https://github.com/wazuh/wazuh/pull/11364) have been carried out on the Wazuh cluster that provides it with multiprocessing capabilities.

This has been a boost in its performance. However, it involves creating new child processes that also use resources. These child processes are not taken into account by this tool, hence the need to update it.

Also, when the tool is updated to aggregate information from all processes, the resource thresholds that were set in the performance test (https://github.com/wazuh/wazuh-qa/issues/1939) will probably fall far short. Therefore, it will be necessary to update it.

- [x] Update CSV parser tool.
- [x] Update cluster performance test.

yanazaeva:
|Related issue|
|---|
|#2468|

## Description

This closes #2468. In this pull, we add a new method in order to sum the parent process values and the children process values. Up to now, the values of the children's processes were not taken into account. Besides, we modified the thresholds accordingly to the new changes implemented in 4.3.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/tools/performance/csv_parser.py
+++ b/deps/wazuh_testing/wazuh_testing/tools/performance/csv_parser.py
@@ -11,6 +11,25 @@ from re import compile
 import numpy as np
 import pandas as pd
 
+aggregation_function = {
+    "Daemon": "first",
+    "Version": "first",
+    "PID": "first",
+    "CPU(%)": "sum",
+    "VMS(KB)": "sum",
+    "RSS(KB)": "sum",
+    "USS(KB)": "sum",
+    "PSS(KB)": "sum",
+    "SWAP(KB)": "sum",
+    "FD": "sum",
+    "Read_Ops": "sum",
+    "Write_Ops": "sum",
+    "Disk_Read(KB)": "sum",
+    "Disk_Written(KB)": "sum",
+    "Disk_Read_Speed(KB/s)": "sum",
+    "Disk_Write_Speed(KB/s)": "sum"
+}
+
 
 class ClusterCSVParser:
     """Class to load and parse CSVs with data produced by the Wazuh cluster.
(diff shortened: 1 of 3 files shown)


deps/wazuh_testing/wazuh_testing/tools/performance/csv_parser.py
```diff
@@ -102,13 +121,37 @@ def _calculate_all_df_stats(self, dfs_dict):
...
         for node, files in dfs_dict.items():
             for phase in [self.SETUP_PHASE, self.STABLE_PHASE]:
-                for file_name, file_df in files.items():
+
+                # Check whether we are calculating the resources' stats
+                for file_df in files.values():
+                    all_files = self._add_child_process_to_parent_process(files) if "Daemon" in file_df else files
+                    break
+
+                for file_name, file_df in all_files.items():
```
Selutario, line 130: Since `phase` is not needed as a parameter for `_add_child_process_to_parent_process`, it is more efficient to call that function outside said loop. Otherwise, it has to calculate the same information multiple times.

On the other hand, it is a bit strange to make a loop with a single iteration to determine if `Default` is inside the dataframe. I would say it's better to use something like `all(...)` and a loop inside. It will check that all dataframes inside `files` contain `Default`, not just the first one. In addition, if the same variable name is used (`files`) instead of a new one (`all_files`), you can avoid this condition:  `if "Daemon" in file_df else files`

```suggestion
            if all("Daemon" in df for df in files.values()):
                # Ensure these dataframes contain resources' stats before trying to add up all children.
                files = self._add_child_process_to_parent_process(files)
            for phase in [self.SETUP_PHASE, self.STABLE_PHASE]:
                for file_name, file_df in files.items():
```

tests/performance/test_cluster/test_cluster_performance/data/10w_50000a_thresholds.yaml
```diff
@@ -3,101 +3,101 @@ tasks:
...
+          mean: 70
+          reg_cof: 2.717
       USS(KB):
         master:
-          max: 309394.8
-          mean: 240663.67742
-          reg_cof: 30
+          max: 204781           # (204 MB)
+          mean: 197162          # (197 MB)
+          reg_cof: 209.6        # (209 MB)
```
Selutario, line 99:
```suggestion
          reg_cof: 209.6
```

tests/performance/test_cluster/test_cluster_performance/data/10w_50000a_thresholds.yaml
```diff
@@ -3,101 +3,101 @@ tasks:
...
         workers:
-          max: 189895.2
-          mean: 117506.60276
-          reg_cof: 30
+          max: 201480           # (201 MB)
+          mean: 145000          # (145 MB)
+          reg_cof: 8224         # (8 MB)
```
Selutario, line 103: This regression coefficient looks fairly high to be in the stable phase. It means that `wazuh-clusterd` process is consuming 8MB more every 5s, it should be close to 0.

Could it be related to a wazuh service restart? Was memory really increasing that much in any worker? In thta case, it could be a memory leak for sure.

tests/performance/test_cluster/test_cluster_performance/data/25w_50000a_thresholds.yaml
```diff
@@ -3,101 +3,101 @@ tasks:
...
+          max: 1069926          # (1 GB)
+          mean: 168979          # (168 MB)
+          reg_cof: 332.64       # (332 B)
         workers:
-          max: 524288           # (0.5 GB)
-          mean: 262144          # (0.25 GB)
-          reg_cof: 1024         # (1 MB)
+          max: 208700           # (200 MB)
+          mean: 116458          # (116 MB)
+          reg_cof: 831          # (831 B)
```
Selutario, line 74:
```suggestion
          reg_cof: 831
```
This is just a personal opinion, but I think I wouldn't add a comment when it comes to bytes (only for MB or GB).

Selutario (changes requested):

Selutario (approved):

juliamagan (approved):

Rebits (approved):
LGTM

PR #2744, opened 2022-03-31
3920

Mayons:
Hello team ! 
I included an enhancement on the add-new-filter feature that includes a Scenario Outline to perform navigation across all the modules of the app.
 A problem we discover developing this test is that we have a problem with the load on the overview page; the module elements are not clickable at the beginning of the load, but a few seconds after the elements turn themselves clickable.
So regarding this problem we discuss with @mauceballos to create a new function on the driver.js file to include a force click with an wait for 1 second to solve this issue temporally. 
We are going to discuss this problem with the development team and tag if it's necessary a new issue to this pr.


tests/frontend/test_kibana_app/cypress/integration/step-definitions/filters/navigate-to-modules.when.js
```diff
@@ -1,13 +1,7 @@
...
-  wazuhMenuButton,
-  modulesButton,
-  securityEventsLink
-} from '../../pageobjects/wazuh-menu/wazuh-menu.page';
-When('The user goes to security-event mod', () => {
-  elementIsVisible(wazuhMenuButton);
-  clickElement(wazuhMenuButton);
-  clickElement(modulesButton);
-  clickElement(securityEventsLink);
+import { forceClickElement, elementIsVisible} from '../../utils/driver';
```
mauceballos, line 2: the elementIsVisible  is imported twice ways

mauceballos (approved):

Issue #3232, opened 2022-09-01
Improvement - Add `packages-dev` support to filebeat and wazuh-app jinja's schemas


roronoasins:
# Description
As we are using `wazuh-ansible` for providing support to some wazuh roles, we found that using the ansible's repo we can't provision `4.4`. These tasks use only the packages repo, so we need to include the customization so we can use the packages-dev as well.

Besides, the framework will add the required fields to the inventory that it uses. https://github.com/wazuh/wazuh-jenkins/issues/3909

# Tasks
- [x] Extend filebeat and dashboard schemas so we can include the `packages-dev` repo

PR #3249, opened 2022-09-02
Improvement - Extend wazuh's jinja schema for filebeat and dashboard

roronoasins:
|Related issue|
|-------------|
|       #3232      |

## Description

This PR aims to extend the jinja schema to be able to use other repos for `filebeat` and `wazuh-app`. This will allow us to install `4.4` as we need a `wazuh-app` version that does not mismatch.

### Updated

- Wazuh's jinja schema with new `filebeat` and `wazuh-app` repo's field


provisioning/roles/wazuh_environment/schema.j2
```diff
@@ -111,6 +111,15 @@ wazuh_custom_packages_installation_indexer_rpm_url: {{role_parameters['wazuh_cus
 {% endif %}
 {% endif %}
 
+
+
+
```
Rebits, line 116: Is required these blank lines?

PR #3271, opened 2022-09-08
Fix tests fails caused by stats socket changes

jmv74211:
|Related issue|
|-------------|
| #3113 |

## Description

This PR aims to fix  the tests that started failing after the changes done in the PR associated to #3113.
The Wazuh/Wazuh PR [#13713](https://github.com/wazuh/wazuh/pull/13713) changed the `analysisd`, `remoted` and `wazuhdb` sockets, which  caused tests to fail, requiring changes in commands passed in and the expected outputs.

### Added

- One new test case for `test_wazuhdb_getconfig.py`

### Updated

- Updated `remoted` socket path 
- All Test cases inputs and outputs in `test_wazuhdb_getconfig.py`
- Updated `test_remoted` modules that required updates in input or output checks
- Update regex to match` expected responses for: `test_get_configuration_sock.py` and `test_rules_verbose.py` from the `test_logtest` suite.
- Update EPS limit tests according to the new changes in `analysisd.state` (now the statistics are reset only when the `wazuh-manager` is restarted)

---


Rebits (approved):
LGTM

Issue #3298, opened 2022-09-15
The integration test `test_integratord_read_json_file_deleted` is failing


Rebits:
## Description

In the nightly for  `4.3` version and  `20220915` revision, Integratord module fails.

</details>

### Current behavior
It seems that `test_integratord_read_json_file_deleted` test fails consistently for `4.3` version (not tested in higher versions)

<details>

<summary> Integratord pytest error output </summary>

```
---------------------------- Captured stdout setup -----------------------------
Restarting wazuh-manager (via systemctl):  [  OK  ]
---------------------------- Captured stderr setup -----------------------------
2022-09-15 00:46:54,768 - wazuh_testing - DEBUG - Set local_internal_option to {'integrator.debug': '2'}
------------------------------ Captured log setup ------------------------------
DEBUG    wazuh_testing:conftest.py:882 Set local_internal_option to {'integrator.debug': '2'}
----------------------------- Captured stderr call -----------------------------
2022-09-15 00:47:19,427 - wazuh_testing - ERROR - Did not recieve the expected "...Could not retrieve information/open file"
2022-09-15 00:47:19,428 - wazuh_testing - ERROR - Results accumulated: 0
2022-09-15 00:47:19,428 - wazuh_testing - ERROR - Results expected: 1
------------------------------ Captured log call -------------------------------
ERROR    wazuh_testing:monitoring.py:458 Did not recieve the expected "...Could not retrieve information/open file"
ERROR    wazuh_testing:monitoring.py:459 Results accumulated: 0
ERROR    wazuh_testing:monitoring.py:461 Results expected: 1
- generated html file: file:///tmp/Test_integration_B31860_20220915013518/report.html -
=========================== short test summary info ============================
FAILED test_integratord/test_integratord_read_json_file_deleted.py::test_integratord_read_json_file_deleted[Cannot read alerts - Json File Deleted]
=================== 1 failed, 4 passed in 109.55s (0:01:49) ====================
```

mauromalara:
## Update 21/09/2022

### Local executions

| OS | R1 | R2 | R3
|-- | -- |-- |--
[...]

PR #3362, opened 2022-09-23
Fix wazuh-integratord tests

mauromalara:
|Related issue|
|-------------|
| #3298 |

## Description

After a change was made in a message, we had to change a callback from ERROR to WARNING.
Also, we have to change the test case IDs to be capable of running tests and selecting them by their IDs.
Finally, we made changes to the current documentation.

Refs: wazuh/wazuh#14431

### Updated

- `test_integratord_change_json_inode`: The sleep time was reduced to 1 second because `wazuh-integratord` has a one-second delay
- `deps/wazuh_testing/wazuh_testing/modules/integratord/__init__.py`: A callback was changed.

---

## Testing performed

| Tester             | Test path | Jenkins | Local  | OS | Commit | Notes                |
|--------------------|-----------|---------|--------|-----|--------|----------------------|
| @mauromalara (Developer)  | test_integratord/ | [🟢](https://github.com/wazuh/wazuh-qa/files/9637004/manager_html_report_Test_integration_B32098_20220923145025-1.zip) [🟢](https://github.com/wazuh/wazuh-qa/files/9637005/manager_html_report_Test_integration_B32098_20220923145025-2.zip) [🟢](https://github.com/wazuh/wazuh-qa/files/9637006/manager_html_report_Test_integration_B32098_20220923145025-3.zip) | [🟢](https://github.com/wazuh/wazuh-qa/files/9637044/R1-manager-ubuntu.zip) [🟢](https://github.com/wazuh/wazuh-qa/files/9637046/R2-manager-ubuntu.zip) [🟢](https://github.com/wazuh/wazuh-qa/files/9637047/R3-manager-ubuntu.zip)
|         |         | Nothing to highlight |
| @user (Reviewer)   |           | ⚫⚫⚫ | :no_entry_sign: :no_entry_sign: :no_entry_sign:  |        |         | Nothing to highlight |

### DIFF
--- a/tests/integration/test_integratord/data/test_cases/cases_integratord_change_inode_alert.yaml
+++ b/tests/integration/test_integratord/data/test_cases/cases_integratord_change_inode_alert.yaml
@@ -1,4 +1,4 @@
-- name: Cannot read alerts - Inode changed
+- name: alerts_file_inode_changed
   description: The alerts.json file inode has changed and it cannot read alerts from it until it reloads.
   configuration_parameters:
     API_KEY: Insert using --integration-api-key parameter
--- a/tests/integration/test_integratord/test_integratord_change_inode_alert.py
+++ b/tests/integration/test_integratord/test_integratord_change_inode_alert.py
@@ -96,7 +96,7 @@ def test_integratord_change_json_inode(configuration, metadata, set_wazuh_config
             brief: Configure the local internal options file.
         - restart_wazuh_function:
             type: fixture
-            brief: Restart wazuh-modulesd daemon before starting a test, and stop it after finishing.
+            brief: Restart all daemons because Integratord depends on multiple daemons. Stop them after finishing.
         - wait_for_start_module:
             type: fixture
             brief: Detect the start of the Integratord module in the ossec.log
@@ -125,16 +125,20 @@ def test_integratord_change_json_inode(configuration, metadata, set_wazuh_config
     remove_file(ALERT_FILE_PATH)
     copy(TEMP_FILE_PATH, ALERT_FILE_PATH)
 
-    # Wait for Inode change to be detected and insert new alert
-    time.sleep(3)
-    run_local_command_returning_output(command)
+    # Wait for Inode change to be detected
+    # The `integratord` library tries to read alerts from the file every 1 second. So, the test waits 1 second + 1
+    # until the file is reloaded.
+    time.sleep(2)
 
     # Monitor Inode Changed
     check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout * 2,
                             callback=callback_generator(integrator.CB_ALERTS_FILE_INODE_CHANGED),
                             error_message=integrator.ERR_MSG_ALERT_INODE_CHANGED_NOT_FOUND)
 
+    # Insert a new alert
+    run_local_command_returning_output(command)
+
     # Read Response in ossec.log
-    check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout,
+    check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout + 2,
                             callback=callback_generator(integrator.CB_PROCESSING_ALERT),
                             error_message=integrator.ERR_MSG_VIRUSTOTAL_ALERT_NOT_DETECTED)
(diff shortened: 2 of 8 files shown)


tests/integration/test_integratord/data/test_cases/cases_integratord_change_inode_alert.yaml
```diff
@@ -1,4 +1,4 @@
-- name: Cannot read alerts - Inode changed
+- name: alerts_file_inode_changed
```
damarisg, line 1: Is there a reason to change each name? Is it a new change in the sintaxis?

tests/integration/test_integratord/test_integratord_change_inode_alert.py
```diff
@@ -96,7 +96,7 @@ def test_integratord_change_json_inode(configuration, metadata, set_wazuh_config
             brief: Configure the local internal options file.
         - restart_wazuh_function:
             type: fixture
-            brief: Restart wazuh-modulesd daemon before starting a test, and stop it after finishing.
+            brief: Restart all daemons because Integratord depends on multiple daemons. Stop them after finishing.
```
damarisg, line 99: What are the multiple daemons you need?

tests/integration/test_integratord/test_integratord_change_inode_alert.py
```diff
@@ -125,16 +125,20 @@ def test_integratord_change_json_inode(configuration, metadata, set_wazuh_config
...
     check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout * 2,
                             callback=callback_generator(integrator.CB_ALERTS_FILE_INODE_CHANGED),
                             error_message=integrator.ERR_MSG_ALERT_INODE_CHANGED_NOT_FOUND)
 
+    # Insert a new alert
+    run_local_command_returning_output(command)
+
     # Read Response in ossec.log
-    check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout,
+    check_integratord_event(file_monitor=wazuh_monitor, timeout=global_parameters.default_timeout + 2,
```
damarisg, line 142: Why did you add `+2` hardcoded?

damarisg (changes requested):

tests/integration/test_integratord/data/test_cases/cases_integratord_change_inode_alert.yaml
mauromalara: @damarisg this is because that name will be used after by pytest to assign an ID to the current test case. So, if we use the name given before my change, we won't be able to run a single test case ([using the `-k` option](https://docs.pytest.org/en/7.1.x/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)).

tests/integration/test_integratord/test_integratord_change_inode_alert.py
mauromalara: @damarisg Because after multiple executions that was the time in which Wazuh managed to process the alert and display the expected message in the log.

tests/integration/test_integratord/test_integratord_change_inode_alert.py
mauromalara: @damarisg as for now, I only consider necessary `wazuh-analisysd` and `wazuh-integratord`; so I will change the restart function to restart only those.

mauromalara:
After modifying the test so that only the necessary daemons are restarted, the test started to run faster.
This causes it to fail and this is because more requests are made per minute than allowed by the VirusTotal API (free).

This message appeared in the `ossec.log`:
```
DEBUG: integratord: Tue Sep 27 15:30:32 -03 2022: 1:virustotal:{"virustotal": {"error": 204, "description": "Error: Public API request rate limit reached"}, "integration": "virustotal"}
```

I see 3 possible solutions:

1. Upgrade to VirusTotal PRO plan.
2. Make another integration that does not require a subscription (ex: Slack).
3. Change the regex that searches the `ossec.log` to: `.*wazuh-integratord.*DEBUG: Command ran successfully.`.

tests/integration/test_integratord/test_integratord_change_inode_alert.py
damarisg: I checked that, and It's correct!

Rebits:
Duplicated https://github.com/wazuh/wazuh-qa/pull/3499

Issue #4552, opened 2023-09-25
System tests fail on agent enrollment


juliamagan:
## Description

During the tests launched in [Release 4.5.3 - RC 1 - System tests](https://github.com/wazuh/wazuh/issues/19118), it has been seen that two tests consistently fail during the agent's enrollment, while waiting for the message:

```
Requesting a key from server:
```

These tests are not exactly the same, but similar, and fail at the same point, so the error could be related. We should investigate what could be happening.

Deblintrake09:
## Research update

- Replicated issue. Logs appear in manager but not on Agent.
<details><summary>Logs evidence</summary>
   - Agent logs
```
2023/09/26 23:53:59 wazuh-agentd[14906] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
2023/09/26 23:53:59 wazuh-agentd[14906] main.c:139 at main(): DEBUG: Wazuh home directory: /var/ossec
2023/09/26 23:53:59 wazuh-agentd[14906] main.c:141 at main(): DEBUG: Started (pid: 14906).
2023/09/26 23:53:59 wazuh-agentd[14906] agent_op.c:219 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
2023/09/26 23:53:59 wazuh-agentd[14906] agent_op.c:238 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [debian, debian10]
2023/09/26 23:53:59 wazuh-agentd[14906] config.c:424 at ReadConfig(): DEBUG: agent_config element does not have any attributes.
2023/09/26 23:53:59 wazuh-agentd[14906] agent_op.c:219 at os_read_agent_profile(): DEBUG: Calling os_read_agent_profile().
2023/09/26 23:53:59 wazuh-agentd[14906] agent_op.c:238 at os_read_agent_profile(): DEBUG: os_read_agent_profile() = [debian, debian10]
2023/09/26 23:53:59 wazuh-agentd[14906] config.c:424 at ReadConfig(): DEBUG: agent_config element does not have any attributes.
2023/09/26 23:53:59 wazuh-execd: INFO: Started (pid: 14921).
2023/09/26 23:54:00 wazuh-agentd[14929] debug_op.c:116 at _log_function(): DEBUG: Logging module auto-initialized
2023/09/26 23:54:00 wazuh-agentd[14929] main.c:139 at main(): DEBUG: Wazuh home directory: /var/ossec
[...]

PR #4562, opened 2023-09-28
Fix enrollment system tests

juliamagan:
|Related issue|
|-------------|
|       #4552       |

## Description

We've found that the test's function `clear_file()` is creating a new file _etc/ossec.log_ with these permissions:

```
rw--r--r-- root root
```

This means that _wazuh-agentd_  is unable to write into _ossec.log_ after dropping its permissions.

### Updated

- Clear `ossec.log` after yield, so the file exists and it isn't created with wrong permissions.

---

## Testing performed

- [🟢 ](https://github.com/wazuh/wazuh-qa/files/12747929/test_enrollment.zip) `test_cluster/test_agent_enrollment`
- [🟢 ](https://github.com/wazuh/wazuh-qa/files/12747931/basic_environment_env.zip) `test_enrollment`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Release report: https://github.com/wazuh/wazuh/issues/19111
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


verdx (approved):
LGTM

PR #4802, opened 2024-01-02
Remove deprecated commands

TomasTurina:
## Description

Removed the deprecated commands `get-groups-integrity`, `select-agent-group` and `find-group` from the tests.

Fix tests that use column `group` and `group_hash` from agent table.

> This PR contains changes from both Framework and the Core team
>
> Linting errors that are not fixed are because of this https://github.com/adrienverge/yamllint/issues/275


Selutario:
Global db will probably undergo major changes in 5.0. Closing since the performance worsens with the changes of this epic.

PR #4822, opened 2024-01-09
Enhancement/4757 dtt1 provision module improvement

fcaffieri:
|Related issue|
|-------------|
|             |


deployability/modules/generic/ansible.py
```diff
@@ -1,55 +1,90 @@
...
+  def set_inventory(self, inventory):
+    """
+    Set the inventory for ansible.
+
+    Args:
+        inventory: Path to the inventory file.
+    """
+    with open(inventory, 'r') as file:
+        inv = yaml.safe_load(file)
+    return inv
```
QU3B1M, line 28: This function only returns the content of an inventory, instead I think this function should save the inventory on the `self.inventory` attribute, so, when we use `set_inventory()` we are modifying the inventory assigned to the class.

```suggestion
    def set_inventory(self, inventory: str | Path) -> None:
        """
        Set the inventory for ansible.
        Args:
            inventory: Path to the inventory file.
        """
        if not Path(inventory).exists():
            raise ValueError(f'Inventory file "{inventory}" does not exist')
        with open(inventory, 'r') as file:
            inventory_data = yaml.safe_load(file)
        self._inventory = inventory_data
```

QU3B1M (changes requested):

deployability/modules/generic/ansible.py
QU3B1M: Anyway, the inventory is not being used anymore, so maybe we can just discard the getters and setters as python doesn't require a getter or setter to access a class attribute or change its value. We can just:
> Imagine this is a dummy version of Ansible where the inventory is just a string
```python
test = Ansible(inventory="inventory_one")
print(f"Inventory: {test.inventory}")
# Inventory: inventory_one

test.inventory = "inventory_two"
print(f"Inventory: {test.inventory}")
# Inventory: inventory_two
```

The getters and setters make sense if we define private attributes (in python thats done with a `_` at the beginning of the attribute or method name). 

I know this function comes from the first version of the Ansible class, but it could be a good time to modify it

QU3B1M (approved):
LGTM

deployability/modules/generic/ansible.py
fcaffieri: Done in [004a0b3](https://github.com/wazuh/wazuh-qa/pull/4822/commits/004a0b39a1bbf2a2c2914f23d88564af5987deed)

deployability/modules/generic/ansible.py
```diff
@@ -1,55 +1,90 @@
...
+from modules.generic.utils import Utils
 
 class Ansible:
 
-    # -------------------------------------
-    #   Variables
-    # -------------------------------------
+  def __init__(self, ansible_data, path=None, inventory=None):
+    self.path = path
+    self.get_inventory = inventory
```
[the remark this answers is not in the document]
fcaffieri: Done in [004a0b3](https://github.com/wazuh/wazuh-qa/pull/4822/commits/004a0b39a1bbf2a2c2914f23d88564af5987deed)

Issue #4846, opened 2024-01-16
DTT1 - Iteration 3 - Test module - Remove usage of Playbook class


QU3B1M:
> Epic #4495

## Description
This issue aims to replace the usage of **Playbook** with the new attributes of the **Ansible** class included in the **Provisioner** at the last iteration
### Tasks
- [x] Remove the Playbook class and all the imports
- [x] Replace `Playbook.PLAYBOOKS_PATH` with `Ansible.playbooks_path`

fcaffieri:
LGTM

PR #4869, opened 2024-01-17
Remove Playbook class and update test module removing its usage

QU3B1M:
## Description

Removes the **`Playbook`** class and its usage inside the Test Module, replaced the `playbook_path` to use the one defined inside the Ansible class.

> Related issue: #4846


fcaffieri (approved):
LGTM

Issue #4839, opened 2024-01-16
FIM Test Fails for Windows Agents


rafabailon:
# Description

During the tests of [Release 4.8.0 - Alpha 2 - Integration tests](https://github.com/wazuh/wazuh/issues/21399) I have detected an error in Windows Agent. The specific error is 

```bash
ERROR    wazuh_testing:monitoring.py:466 TimeoutError was raised because a single 'modified' event was expected for ['c:\\testdirectory_5\\testfile0'] but was not detected.
```

The error has been detected in the FIM tests.

<details>
  <summary>Pipeline Error</summary>
  <br/>

```bash
11:59:33  TASK [Launch windows pytests] **************************************************
11:59:33  task path: /home/ec2-user/workspace/Test_integration/quality/tests/integration/test_integration.yaml:37
11:59:33  fatal: [ID2_B45851_Test_integration_windows_agent]: FAILED! => {
11:59:33      "ansible_async_watchdog_pid": 2808,
11:59:33      "ansible_job_id": "905546570871.1016",
11:59:33      "changed": true,
11:59:33      "cmd": "powershell.exe -",
11:59:33      "delta": "0:25:48.621535",
11:59:33      "end": "2024-01-15 10:58:40.488711",
11:59:33      "failed_when_result": true,
11:59:33      "finished": 1,
11:59:33      "rc": 1,
11:59:33      "results_file": "C:\\Users\\Jenkins\\.ansible_async\\905546570871.1016",
11:59:33      "start": "2024-01-15 10:32:51.867175",
11:59:33      "started": 1
11:59:33  }
11:59:33  
11:59:33  MSG:
11:59:33  
11:59:33  non-zero return code
11:59:33  Using module file /usr/local/lib/python3.7/site-packages/ansible/modules/windows/win_stat.ps1
11:59:33  Pipelining is enabled.
11:59:33  <172.31.4.162> ESTABLISH WINRM CONNECTION FOR USER: Jenkins on PORT 5986 TO 172.31.4.162
11:59:33  EXEC (via pipeline wrapper)
11:59:37  <172.31.4.162> FETCH "C:\tmp\Test_integration_B45851_20240115110601\pytest_plain_report.txt" TO "/home/ec2-user/workspace/Test_integration/windows_agent_tests_Test_integration_B45851_20240115110601.txt"
...
11:59:45  PLAY RECAP *********************************************************************
[...]

PR #4873, opened 2024-01-18
Fix test_file_checker in check_mtime case

jotacarma90:
|Related issue|
|-------------|
|https://github.com/wazuh/wazuh-qa/issues/4839|

## Description

This PR is to fix some false negative cases that were coming up in the `test_file_checkers`.

This test configures a Wazuh directory with `check_all=no` and `check_mtime=yes` so only the `timestamp` will be taken into account by FIM to see if there have been changes in the file. And if it coincides that the added and the modified happen in the same second (same timestamp), the modification event is not detected and that's why the test fails.

To solve it I have included a sleep of one second after the creation of the file for the case of the `check_mtime`. This way we make sure that the timestamp of the event is different and we get correctly the alert.

Jenkins build success:
https://ci.wazuh.info/job/Test_integration/45871/

### DIFF
--- a/tests/integration/test_fim/test_files/test_file_checks/test_file_checkers.py
+++ b/tests/integration/test_fim/test_files/test_file_checks/test_file_checkers.py
@@ -71,6 +71,7 @@ from wazuh_testing.tools.configuration import load_wazuh_configurations, check_a
 from wazuh_testing.tools.monitoring import FileMonitor
 from wazuh_testing.modules.fim.utils import regular_file_cud
 from wazuh_testing.tools import PREFIX
+from time import sleep
 
 # Marks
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_fim/test_files/test_file_checks/test_file_checkers.py
juliamagan, line 1: Why not simply add a sleep inside the function that modifies the files?

juliamagan (changes requested):

juliamagan (approved):
The `regular_file_cud` function does too many things, it is practically the whole test. The best solution would be to refactor this function to be able to split the different tasks, but this would affect most of the FIM tests, and knowing that these tests will be deprecated in the future and worked on by the core team to implement them in GitHub Actions, we can accept this solution.

tests/integration/test_fim/test_files/test_file_checks/test_file_checkers.py
```diff
@@ -207,6 +208,14 @@ def test_checkers(file_path, file_attrs, tags_to_apply, triggers_modification, c
...
     check_apply_test(tags_to_apply, get_configuration['tags'])
 
+    def waitasecond(event):
+        sleep(1)
+
     # Test files checks.
-    regular_file_cud(file_path, wazuh_log_monitor, min_timeout=global_parameters.default_timeout,
-                     options=file_attrs, triggers_modified_event=triggers_modification, escaped=True)
+    if file_attrs == {CHECK_MTIME}:
+        regular_file_cud(file_path, wazuh_log_monitor, min_timeout=global_parameters.default_timeout,
```
lchico, line 216: Could you please add a comment explaining briefly why we are doing this?

lchico (changes requested):
I left a comment, It would be helpful to have a comment in the code explaining why we are doing this.

lchico (dismissed):
LGTM!

lchico (approved):
LGTM

MarcelKemp (approved):

juliamagan (approved):

PR #4904, opened 2024-02-02
Update schema argument to be optional

QU3B1M:
# Description

This PR sets a default `schema.json` to be used in the execution with the possibility to use a custom schema validator..


fcaffieri (approved):

pro-akim (approved):
## Review Notes
LGTM

Issue #4854, opened 2024-01-16
DTT1 - Iteration 3 - Allocation module - Improve sizes and OS for AWS provider


QU3B1M:
> Epic #4855

## Description
This issue aims to add new sizes and operating systems for the AWS provider, including validating the correct functioning of these OSs along with those added previously.

### Tasks
- [x] Add required OS
  - [x] RedHat	7
  - [x] RedHat	8
  - [x] RedHat	9
  - [x] CentOS	7
  - [x] CentOS	8
  - [x] Debian	10
  - [x] Debian	11
  - [x] Debian	12
  - [x] Ubuntu	18
  - [x] Ubuntu	20
  - [x] Ubuntu	22
  - [x] Oracle Linux
  - [x] Amazon Linux
  - [x] Amazon Linux
  - [x] openSUSE	15
  - [x] SUSE	15
  - [x] Fedora	37
  - [x] Fedora	38
  - [x] Windows	10
  - [x] Windows	11
  - [x] Windows	Server 2012
  - [x] Windows	Server 2012 R2
  - [x] Windows	Server 2016
  - [x] Windows	Server 2019
  - [x] Windows	Server 2022
- [x] Add required sizes

c-bordon:
### Ubuntu 22.04 amd64:

```console
cbordon@cbordon-MS-7C88:~/Documents/wazuh/repositorios/wazuh-qa/deployability$ python3 launchers/allocation.py --composite-name linux-ubuntu-22.04-amd64 --provider aws --size large
[2024-02-07 12:30:20] [INFO] ALLOCATOR: Creating instance at /tmp/wazuh-qa
[2024-02-07 12:30:20] [DEBUG] ALLOCATOR: No config provided. Generating from payload
[2024-02-07 12:30:20] [DEBUG] ALLOCATOR: Creating base directory: /tmp/wazuh-qa/AWS-295E17C9-EB9B-43F8-A201-5F0EBAA68CB2
[2024-02-07 12:30:39] [DEBUG] ALLOCATOR: Renaming temp /tmp/wazuh-qa/AWS-295E17C9-EB9B-43F8-A201-5F0EBAA68CB2 directory to /tmp/wazuh-qa/i-0c4cb703324928236
[2024-02-07 12:30:39] [INFO] ALLOCATOR: Instance i-0c4cb703324928236 created.
[2024-02-07 12:30:41] [INFO] ALLOCATOR: Instance i-0c4cb703324928236 started.
[...]

PR #4932, opened 2024-02-07
Updated AWS amis

c-bordon:
Close https://github.com/wazuh/wazuh-qa/issues/4854
Updated the AMIs for AWS deployment:

### Windows server 2022: https://github.com/wazuh/wazuh-qa/issues/4854#issuecomment-1932688109

---

we had to modify this code:

```console
        arch = params.composite_name.split('-')[-1]
        # Parse the configuration.
        for spec in size_specs:
            if fnmatch.fnmatch(arch, spec):
                config['type'] = size_specs[spec]['type']
                break

```
This is because it was limited only to Linux systems, and it was necessary to add it for macOS and Windows, this the AMI is now searched by architecture.

We also had to update the amd64 micro and small instances since they were not compatible with Windows 11 systems onwards.

### DIFF
--- a/deployability/modules/allocation/allocation.py
+++ b/deployability/modules/allocation/allocation.py
@@ -42,7 +42,7 @@ class Allocator:
 
         Args:
             payload (CreationPayload): The payload containing the parameters
-                                       for instance creation.
+                                        for instance creation.
         """
         instance_params = models.CreationPayload(**dict(payload))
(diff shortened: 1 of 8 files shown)


teddytpc1 (approved):

QU3B1M (approved):
LGTM!

Issue #4931, opened 2024-02-07
Filebeat provisioning role includes filebeat template from production only


Rebits:
### Description

During the investigation of [Issue #21789](https://github.com/wazuh/wazuh/issues/21789), it was discovered that the filebeat module consistently retrieves the filebeat template from the production environment. However, in the case of providing staging or pre-release packages, it is crucial to use the appropriate template from the staging environment.

To address this issue, the filebeat module should fetch the template from one of the following URLs when staging or pre-release packages are being utilized:
- [Pre-release Template URL](https://packages-dev.wazuh.com/pre-release/filebeat/wazuh-filebeat-0.4.tar.gz)
- [Staging Template URL](https://packages-dev.wazuh.com/staging/filebeat/wazuh-filebeat-0.4.tar.gz)

> [!NOTE]
> The Filebeat 0.4 templates were originally created as per the instructions outlined in [this internal development request](https://github.com/wazuh/internal-devel-requests/issues/596#issuecomment-1860537979).

MARCOSD4:
## Update

The PR with the changes has been created: https://github.com/wazuh/wazuh-qa/pull/4950.
The pipeline has been tested [here](https://ci.wazuh.info/job/Wazuh_QA_environment/928/) but it failed, we need to investigate why and make the appropriate changes.

MARCOSD4:
## Update

After a meeting with @Rebits some modifications has been made to the code but the pipeline has failed again [here](https://ci.wazuh.info/job/Wazuh_QA_environment/936/) and has to be re-investigated.

Rebits:
Adjusting the ETA to accommodate the final review process.

davidjiglesias:
LGTM

PR #4950, opened 2024-02-12
Add pre-release and staging Filebeat module template URL

MARCOSD4:
# Description

This PR aims to add the corresponding URLs for the filebeat module template when staging or pre-release packages are being utilized since it was discovered that the filebeat module consistently retrieves the filebeat template from the production environment.

---

Related issue: https://github.com/wazuh/wazuh-qa/issues/4931


juliamagan (changes requested):
GJ! Please add an entry with this change in the changelog.

juliamagan (changes requested):
I think it should go in `Fixed` rather than `Changed`. Also, the issue is reported as a bug. Note also that we are filling the changelog from the bottom up.

juliamagan (approved):
LGTM

Issue #4977, opened 2024-02-19
FIM performance test: Obtain database size


juliamagan:
## Description

During the tests performed in https://github.com/wazuh/wazuh-qa/issues/4966, it has been seen that it can be interesting to see how the size of the agent database varies during the execution of the test.

It is proposed to add the necessary methods to obtain this information.

PR #4978, opened 2024-02-19
Obtain database size

juliamagan:
|Related issue |
|:--:|
| #4977 |

## Description

A function has been added to get the size of the database in bytes during the test and store it in a CSV.


Rebits (approved):
LGTM

PR #5155, opened 2024-03-26
Fix in `remote_operations_handler.py` functions 

MARCOSD4:
# Description

This PR aims to make the necessary changes in the code to avoid that the Vulnerability Detector E2E tests do not continuing if any error occurs during the setup of some of the agents, that is to say, if any error occurs during this phase, the execution must continue to allow the development of the rest of the agents. For this purpose, some functions of the `remote_operations_handler.py` script have been modified, since the problem was detected in them because, in the event of an error, they terminated the execution of the function using `raise error`. They have been modified so that they do not terminate the execution but simply report the error and continue the process.

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/end_to_end/remote_operations_handler.py
+++ b/deps/wazuh_testing/wazuh_testing/end_to_end/remote_operations_handler.py
@@ -164,37 +164,43 @@ def install_package(host: str, operation_data: Dict[str, Dict], host_manager: Ho
     package_id = None
 
     if host_os_name in install_package_data:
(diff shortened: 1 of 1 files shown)


Rebits (approved):
Good job overall. However, there are a couple of issues that need to be addressed:

- The current approach doesn't prevent the test case from running for endpoints that failed in the setup phase, leading to inefficient testing.
- It's noted that these functions will undergo refactoring in the future to address code coupling.

Despite these concerns, we'll proceed with these changes for now. This will enable us to generate a comprehensive report for all endpoints, even in failure cases

PR #5234, opened 2024-04-17
Add vulnerabilities and change packages in some Windows upgrade cases

MARCOSD4:
## Description

This PR aims to add some vulnerabilities in the `vuln_packages.json` file that were not contemplated before and add some new packages not vulnerable in some Windows upgrade cases. The `test_vulnerability.yaml` file has also been modified to incorporate these changes.

## Testing performed

These changes have been tested locally by running only the cases that have been modified, that is, from the `upgrade_package_maintain_add_vulnerability0` case onwards. The tests have failed, but in the alert file, you can see the alerts corresponding to the vulnerabilities of each installed package, so the test performance is as expected. The errors are due to the tests failing to detect vulnerabilities because of a change in the index reported in https://github.com/wazuh/wazuh-qa/issues/5239. 

Report: [report.zip](https://github.com/wazuh/wazuh-qa/files/15012741/report.zip)

### DIFF
--- a/deps/wazuh_testing/wazuh_testing/end_to_end/vulnerability_detector_packages/vuln_packages.json
+++ b/deps/wazuh_testing/wazuh_testing/end_to_end/vulnerability_detector_packages/vuln_packages.json
@@ -455,6 +455,8 @@
     "package_name": "node",
     "package_version": "18.0.0",
     "CVE": [
+      "CVE-2023-44487",
+      "CVE-2023-23936",
       "CVE-2023-38552",
       "CVE-2023-32559",
       "CVE-2023-32006",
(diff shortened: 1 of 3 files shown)


deps/wazuh_testing/wazuh_testing/end_to_end/vulnerability_detector_packages/vuln_packages.json
juliamagan, line 1: If we are not going to use the 19.5.0 or 19.6.0 packages anymore, we can delete them.

juliamagan (changes requested):
GJ! Remember to add the changelog entry

deps/wazuh_testing/wazuh_testing/end_to_end/vulnerability_detector_packages/vuln_packages.json
MARCOSD4: I have not deleted them in case they are needed in the future, there are also other packages that are not used in the file.

Issue #5345, opened 2024-05-07
Change manager Vulnerability Detector E2E test case logs are truncated


Rebits:
# Description

It has been detected in [Release 4.8.0 - RC 1 - Vulnerability Detection E2E tests](https://github.com/wazuh/wazuh/issues/23299), that the test case `test_change_agent_manager` is truncating the agent/manager logs, making difficult to get any information of the environment in case of failure.

santipadilla:
Investigating the truncation process, we can see that:

1.  clean_environment_logs and clean_environment_logs_function: both functions perform similar actions but have a different scope. They manage logs by deleting them at the end of their scope using truncate_remote_host_group_files.

2. truncate_remote_host_group_files: This function directly handles log deletion. It takes parameters to decide which host group and file types should be truncated. 

3. truncate_file: This function performs the actual action of opening the specified file in write mode, which automatically truncates the file.

santipadilla:
### Test before changes

- Truncating logs for the `test_change_agent_manager` case

[report.zip](https://github.com/wazuh/wazuh-qa/files/15260073/report.zip)

santipadilla:
Moved ETA due to final review

Rebits:
During a meeting with @santipadilla, we identified that the new agent configuration isn't properly applied after the configured agent is permuted because the agent isn't restarted following the change. It's likely that the restart agent call was inadvertently omitted during a merge or PR review.

The solution appears straightforward: ensure that agents are restarted after the `permutate_manager` fixture is applied.

> [!WARNING]
> It's imperative to rerun this test in the E2E environment. Unfortunately, a current bug in RC1 (https://github.com/wazuh/wazuh/issues/23301) is hindering the appearance of some vulnerabilities. Therefore, we'll need to wait until this issue is resolved before verifying if there are any product-related issues concerning the manager change.

santipadilla:
## Update

### Test before modifying `permutate_agents_managers` fixture

[report.zip](https://github.com/wazuh/wazuh-qa/files/15292722/report.zip)

santipadilla:
New changes made in this branch: `fix/5345-skip-truncate-test-change-agent-manager`
In this PR: https://github.com/wazuh/wazuh-qa/pull/5355

PR #5355, opened 2024-05-09
Agents are not being restarted after manager change in E2E vulnerability detection tests

santipadilla:
|Related issue|
|-------------|
|    #5345         |

## Description

This PR fix the restart of the agents in  `test_change_agent_manager` case.

### DIFF
--- a/tests/end_to_end/test_vulnerability_detector/conftest.py
+++ b/tests/end_to_end/test_vulnerability_detector/conftest.py
@@ -139,7 +139,7 @@ def clean_environment_logs(host_manager: HostManager, request):
     skip_truncate_tests = ['test_change_agent_manager']
 
     if any(test in request.node.name for test in skip_truncate_tests):
-        print("Skipping truncation of logs for this test.")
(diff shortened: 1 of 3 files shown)


rafabailon (approved):
LGTM

tests/end_to_end/test_vulnerability_detector/conftest.py
juliamagan, line 1: I don't think this is the best approach. It may be necessary that the logs are truncated, what we must avoid is that they are not truncated too much. For example, the affected test uses two fixtures to truncate the logs.

juliamagan (changes requested):

Rebits (approved):
LGTM

tests/end_to_end/test_vulnerability_detector/conftest.py
santipadilla: Fixed here: https://github.com/wazuh/wazuh-qa/pull/5355/commits/d99ede4bfa859316019fa7a820932d448ab8e8f5

juliamagan (approved):

Issue #5367, opened 2024-05-13
DTT1 - Iteration 3 - Allocation module - Expose the port of the macOS Intel VMs in the same way as it is done on the ARM host


c-bordon:
## Description
The objective of this issue is to expose the port for Intel machines on the macStadium host in the same way we do with ARM. The reason is to ensure that access to the VM is through the private IP only. We have tested this in this issue https://github.com/wazuh/wazuh-qa/issues/5354 and we had different behaviors depending on the virtualizer, in the case of VirtualBox the configuration worked correctly, but in the case of Parallels, despite having the 'host_ip' configuration correctly configured, the VM could be accessed through both the public IP and the private IP.

- [x] Use the same port forwarding over SSH method for Intel machines
- [x] Unify the validation method of the inventory.yaml and track.yaml files as done in this branch: https://github.com/wazuh/wazuh-qa/tree/change/5354-dtt1-iteration-3-allocation-module-configure-vms-created-in-macstadium-with-custom-password-or-ssh-key

c-bordon:
## macOS Intel test

```console
cbordon@cbordon-MS-7C88:~/Documents/wazuh/repositorios/wazuh-qa$ python3 deployability/modules/allocation/main.py --provider vagrant --size small --instance-name cbordon-test --composite-name macos-sonoma-14-amd64
[2024-05-14 16:31:50] [INFO] ALLOCATOR: Creating instance at /tmp/wazuh-qa
[2024-05-14 16:32:03] [INFO] ALLOCATOR: Using the macStadium Intel server to deploy.
[2024-05-14 16:32:03] [DEBUG] ALLOCATOR: Checking if instance directory exists on remote host
[2024-05-14 16:32:06] [DEBUG] ALLOCATOR: Creating instance directory on remote host
[2024-05-14 16:32:09] [DEBUG] ALLOCATOR: No config provided. Generating from payload
[2024-05-14 16:32:09] [DEBUG] ALLOCATOR: Generating new key pair
[2024-05-14 16:32:12] [DEBUG] ALLOCATOR: Vagrantfile created. Creating instance.
[2024-05-14 16:32:24] [INFO] ALLOCATOR: Instance cbordon-test-9914 created.
[2024-05-14 16:33:11] [INFO] ALLOCATOR: Instance cbordon-test-9914 started.
[...]

PR #5389, opened 2024-05-15
Validation of inventory and track files is improved, and private IP access is configured for macOS

c-bordon:
close https://github.com/wazuh/wazuh-qa/issues/5367

Improvements are made to the validations of the inventory and track files, allowing more configuration options and reducing the possibility of errors.

#### Testing path with track word: https://github.com/wazuh/wazuh-qa/issues/5367#issuecomment-2115321403

We also make the necessary configurations so that macOS VMs can only be accessed by the private IP. With the previous configuration, the Intel macOS machines could access the VMs through the private or public IP. We had to make adaptations to make this work in both virtualizers (VirtualBox and Parallels)

#### macOS Intel Parallels test: https://github.com/wazuh/wazuh-qa/issues/5367#issuecomment-2111004815

### DIFF
--- a/deployability/modules/allocation/allocation.py
+++ b/deployability/modules/allocation/allocation.py
@@ -2,11 +2,14 @@
 # Created by Wazuh, Inc. <info@wazuh.com>.
 # This program is a free software; you can redistribute it and/or modify it under the terms of GPLv2
 
-import yaml
-import paramiko, logging, time
-import winrm
-
+import time
+import os
 from pathlib import Path
+
+import yaml
+import paramiko
+import logging
+import winrm
 
 from .aws.provider import AWSProvider, AWSConfig
 from .generic import Instance, Provider, models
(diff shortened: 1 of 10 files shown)


deployability/modules/allocation/allocation.py
```diff
@@ -258,3 +259,33 @@ def __check_connection(inventory: models.InventoryOutput, attempts=30, sleep=30)
...
+            return Path(os.path.join(instance.path, file + '.yml'))
+        if path.exists() and path.is_dir():
+            return Path(os.path.join(path, file + '.yml'))
+        if str(path).endswith('.yml') or str(path).endswith('.yaml'):
+            if file in str(path):
+                return path
+            elif file == 'inventory' and 'track' in str(path):
+                return Path(str(path).replace('track', 'inventory'))
+            elif file == 'track' and 'inventory' in str(path):
+                return Path(str(path).replace('inventory', 'track'))
```
rauldpm, line 286: I don't think we should assume the error and auto-fix it if we validate and force that those files should have a specific name, then we can work with this, but we need to be sure not to accept false positives

We can also add an ID like `type: 'inventory'` in the inventory and track files, this will avoid checking names and content

rauldpm (changes requested):

rauldpm (approved):
LGTM

teddytpc1 (approved):

rauldpm (approved):

PR #5499, opened 2024-06-14
Update wazuh_template_branch to 4.8.1

Rebits:
# Description

This PR update wazuh_template_branch to 4.8.1

## Testing performed

Build:  https://ci.wazuh.info/job/Test_e2e_system/301/

### DIFF
--- a/provisioning/roles/wazuh/ansible-filebeat-oss/defaults/main.yml
+++ b/provisioning/roles/wazuh/ansible-filebeat-oss/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
 filebeat_version: 7.10.2
 
-wazuh_template_branch: 4.8.0
+wazuh_template_branch: 4.8.1
 
 filebeat_node_name: node-1
 


santipadilla (approved):

provisioning/roles/wazuh/ansible-filebeat-oss/defaults/main.yml
juliamagan, line 1: Checking the repository we could see that there are more files with the hardcoded version, for example, `provisioning/roles/wazuh/ansible-wazuh-agent/defaults/main.yml`:


https://github.com/wazuh/wazuh-qa/blob/77da068720e4eaadb9ccd98884d1fc470f8f7275/provisioning/roles/wazuh/ansible-wazuh-agent/defaults/main.yml#L3


In addition, this is not maintainable, we must check if this version is essential (in the same playbook in the installation by sources is v4.4.4) and if it is, look for a way to parameterize it, if not, we will have to take into account these files every time we make a version bump.

juliamagan (changes requested):

provisioning/roles/wazuh/ansible-filebeat-oss/defaults/main.yml
Rebits: These values are default and not currently used in provisioning.

The inclusion of some playbooks from a wazuh-ansible repository for provisioning the E2E VD environment was a condition accepted. This has resulted in non-maintainable code. However, refactoring is not planned as we intend to migrate to new tools in the future.

I suggest replacing all hardcoded values in each version bump

Rebits: Proposed changes will be carried on https://github.com/wazuh/wazuh-qa/issues/5509

juliamagan (approved):

PR #5712, opened 2024-09-03
Increase results windows in E2E Vulnerability detection

### LINKED ISSUE #5699: Increase E2E vulnerability detection scans timeout
### Description

In [Release 4.9.0 - RC 1 - Vulnerability Detection E2E tests](https://github.com/wazuh/wazuh/issues/25363) it was detected several errors:
- Inconsistent results between initial scan tests
- Some vulnerabilities were detected, generating expected alerts but vulnerabilities were not correctly indexed. 

After research (https://github.com/wazuh/wazuh/issues/25363#issuecomment-2309806230) it was concluded that these errors were motivated due to a regression in the times of the indexer and for this reason it's necessary to increase the timeout of the initial tests scans and the timeout for collecting vulnerabilities in the syscollector case for each agent.

### Tasks
- [x] ~Increase vulnerability timeouts for initial scans and package syscollector cases.~ Increase result windows 

### Validation

- [x] Launch a https://ci.wazuh.info/job/Test_e2e_system/E2E build were tests no longer fail due to this issue: https://github.com/wazuh/wazuh-qa/issues/5699#issuecomment-2326993520

### Conclusion

https://github.com/wazuh/wazuh-qa/issues/5699#issuecomment-2326993520

Rebits:
# Description

IT was detected in https://github.com/wazuh/wazuh-qa/issues/5699, that the E2E Vulnerability detection tests were failing due to the result windows for the Indexer were not enough for some agents in the environment, leading to some vulnerabilities being missing.

More information at https://github.com/wazuh/wazuh-qa/issues/5699#issuecomment-2326993520

# Changes
- Increase result windows to 50000

## Testing performed

**Build**: https://ci.wazuh.info/job/Test_e2e_system/368/

> [!NOTE]
> Tested along with https://github.com/wazuh/wazuh-qa/issues/5699 in a temporal branch

### DIFF
--- a/tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
+++ b/tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
@@ -70,7 +70,9 @@ from wazuh_testing.end_to_end.vulnerability_detector import (TIMEOUT_PER_AGENT_V
                                                              get_vulnerabilities_from_states_by_agent)
 from wazuh_testing.end_to_end.waiters import wait_until_vd_is_updated
 from wazuh_testing.tools.system import HostManager
-
+from wazuh_testing.end_to_end.indexer_api import (INDEXER_RESULT_WINDOWS_VULN_E2E,
+                                                 extend_result_window,
+                                                 get_wazuh_states_vulnerabilities_indexname)
 
 pytestmark = [pytest.mark.e2e, pytest.mark.vulnerability_detector, pytest.mark.tier0]
 
@@ -192,6 +194,11 @@ def configure_vulnerability_detection_test_environment(
     yield test_timestamp
 
 
+def max_result_window(host_manager):
+    vuln_index = get_wazuh_states_vulnerabilities_indexname()
+    extend_result_window(host_manager, index=vuln_index, new_max_result_window=INDEXER_RESULT_WINDOWS_VULN_E2E)
+
+
 @pytest.mark.filterwarnings("ignore::urllib3.exceptions.InsecureRequestWarning")
 class TestInitialScans:
     # Checks definition
@@ -254,7 +261,7 @@ class TestInitialScans:
         configure_vulnerability_detection_test_environment,
         record_property,
         clean_environment_logs,
-        delete_states_vulnerability_index,
+        delete_states_vulnerability_index
     ):
         """
         description: Validates the initiation of the first Syscollector scans across all agents in the environment.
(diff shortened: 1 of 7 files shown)


tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
```diff
@@ -501,10 +510,12 @@ def test_syscollector_second_scan(
 
         global FIRST_SCAN_TIME
 
+        max_result_window(host_manager)
```
jseg380, line 513: I can't help but notice max_result_window is called several times yet the variable host_manager **doesn't seem** to be updated in any way (assignment in this file, modification in some function/method where it is passed as a parameter). Is this intentional?
If the reason to do this is increase several times the result window I believe it could be replaced by a single bigger increment.

tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
```diff
@@ -928,8 +943,10 @@ def test_vulnerability_detector_scans_cases(self, request, preconditions, body,
             package_data = [body["package"]["to"], body["package"]["from"]]
         else:
             package_data = [body["package"]]
+        max_result_window(host_manager)
```
jseg380, line 946: Most of the calls to `max_result_window` are done prior to calling the function `get_vulnerabilities_index`

tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
jseg380: Except for this one that is called before `get_vulnerabilites_from_states_by_agent`
**Not sure**, but if it's always called with these functions it could be either added to those functions or create a wrapper function to call both functions together in the correct order.

Rebits: This approach was intentional. While it isn't the cleanest solution, several limitations have led us to adopt it:

- Extending the result window requires the index to exist, which rules out fixtures that perform this operation at the start of the test since the index is frequently deleted during the initial scans.
- The tests are designed to be run independently, so making changes in only one test could result in unpredictable behavior that's difficult to debug.
 - Although it may be possible to improve this logic by refactoring the tests, these tests will be completely revamped for version 5.0.0. Therefore, I recommend not investing time in this for now.

tests/end_to_end/test_vulnerability_detector/test_vulnerability_detector.py
Rebits: Check https://github.com/wazuh/wazuh-qa/pull/5712#discussion_r1745290989. In addition I prefer explicitly include this operation in the test rather than include not expected logic in `get_vulnerabilities_index`

jseg380 (approved):
LGTM

juliamagan (approved):
LGTM

PR #5803, opened 2024-10-08
Adding macOS Sequia support

pro-akim:
# Description

Adding macOS Sequoia 15 to DTT tests


rauldpm (changes requested):
Typo in commit (sequia) (enhacement)

